PR #20, opened 2023-04-12
Updates and fixes for the existing Docker-based hosting capability

sfpurdy:
- `.env-sample`
  - Added external port and admin user variables
  - Moved comments to above the relevant line to prevent issues with the `env_file` directive in Docker Compose
- `docker-compose.override.yml`
  - Added
  - The override file is automatically used when running locally so it doesn't impact someone running with `docker-compose up`, but this makes it easier to use a named volume in a Docker Swarm environment
- `docker-compose.swarm.yml`
  - Added for running in a Docker Swarm Environment
  - Adds an environment variable for the built images so you can use a local Docker Registry
    - Also adds a tag for the images. `CI_COMMIT_SHA` is the Git Commit SHA and is automatically added in a Gitlab CI/CD environment
  - Added another volume for automated PostgreSQL backups
- `docker-compose.yml`
  - Move the volumes for Bayanat to the `docker-compose.override.yml` file
  - Use a named volume for the redis service
    - Removed the `rw` as that's the default
  - `bayanat` service - Set the PostgreSQL-related environment variables as well as the credentials for the initial admin user - Use `env_file` to set the environment variables
  - `celery` service
    - Set the PostgreSQL-related environment variables
    - Use `env_file` to set the environment variables
  - `postgres` service - Removed the `env_file` in favor of setting the environment variable in the `environment` section. Dokcer Compose will read the values from the `.env` file if it is there, otherwise it will use the defaults which match the values in `.env-sample`
      - This makes sure the username and password are set so the Postgres container will start up - Also added the environment section with the PostgreSQL variables to the `celery` and `bayanat` services to make sure they match
    - Added default values for the database (`-d`), and username (`-U`) arguments for the healthcheck command as it was failing when these weren't set, causing the service to show as unhealthy
  - Removed the `ENV_FILE` build argument from the `celery` and `bayanat` services in favor of the setting that at runtime using `env_file` - Copying in the `.env` file during the build is problematic as it means you have to rebuild the container to change a configuration or at least mount a new `.env` file at runtime. The `python-dotenv` package will use environment variables instead of the `.env` files anyway, so it's easier to just set them in the `env_file` section.
  - Ports
    - Removed the reference to `127.0.0.1` in the port settings as it isn't necessary and may cause issues on remote servers
    - Add environment variables for the services that expose a port so it can be easily changed
      - These will default to the current port
  - Removed the `container_name` settings on all services as this can cause problems if running multiple versions of the Bayanat system at the same time. The container names will default to `<folder name>_<service name>`, i.e. the default name for the `redis` service if the folder is named `bayanat` will be `bayanat_redis`.
- `enferno/commands.py`
  - Updates the `install` command to allow the username and password to be passed in for unattended installations using environment variables
- `enferno/settings.py`
  - See `PostgreSQL Connection Dropping` item below
- `enferno/tasks/__init__.py`
  - See `PostgreSQL Connection Dropping` item below
- `flask/bin/entrypoint.sh`
  - Create an admin user as a part of the `flask` role startup
- `flask/Dockerfile`
  - Use the `python3.8-` versions for the `python3-dev` and the `python3.10-venv` packages as `python3.10-venv` doesn't seem to exist in the apt repo and it makes sense to specify a particular version to avoid issues later if the default version updates
  - Removed the `ENV_FILE` build arg and `COPY` command in favor of using environment variables set at runtime using docker-compose (see note in `docker-compose.yml`)
  - Add the `ADMIN_USERNAME` and `ADMIN_PASSWORD` environment variables
- `flask/bin/entrypoint.sh`
  - Add the credentials for the initial admin user and pass them to the `flask install` command

### PostgreSQL Connection Dropping

There was an error that caused the PostgreSQL connection to drop when the cached connection goes stale. The fix to the error required the `pool_pre_ping=True` configuration to be passed to SQL Alchemy per the [SQLAlchemy Documentation](https://docs.sqlalchemy.org/en/14/core/pooling.html#pool-disconnects). This fix uses the `SQLALCHEMY_ENGINE_OPTIONS` option of the `flask-sqlalchemy` module to set the `pool_pre_ping` configuration as outlined in the [Module Docs](https://flask-sqlalchemy.palletsprojects.com/en/2.x/api/#flask_sqlalchemy.SQLAlchemy)

- `enferno/tasks/__init__.py`
- Added `celery.conf.update({'database_engine_options': cfg.SQLALCHEMY_ENGINE_OPTIONS})` to set the value

- `enferno/settings.py`
- Added a config for `SQLALCHEMY_ENGINE_OPTIONS` that defaults to `True`

### DIFF
--- a/.env-sample
+++ b/.env-sample
@@ -1,28 +1,39 @@
 # Bayanat
 FLASK_APP=run.py
-FLASK_DEBUG=0 # must be always set to 0 in production deployments
+# FLASK_DEBUG must be always set to 0 in production deployments
+FLASK_DEBUG=0
 
-POSTGRES_DB=bayanat # should leave as default
+# Should leave POSTGRES_DB as the default
+POSTGRES_DB=bayanat
 
 # Uncomment the following lines if you're deploying with Docker
 # or need to supply these information, e.g. postgres or redis are
 # installed on a separate host or require authentication.
 # Otherwise you can leave commented.
 
-# POSTGRES_USER= # should leave as default
-# POSTGRES_PASSWORD= # generate secure password and setup psql accordinngly 
-# POSTGRES_HOST= # use postgres for docker, otherwise add ip/domain name
-# REDIS_HOST= # same as above, redis for docker, or ip/domain name
-# REDIS_PASSWORD= # generate secure password and setup redis accordinngly 
-# REDIS_AOF_ENABLED=no 
-# MEDIA_PATH= # use to point to 
+# Should leave POSTGRES_USER as default
+# POSTGRES_USER=postgres
+# Generate secure password for POSTGRES_PASSWORD and setup psql accordinngly 
+# POSTGRES_PASSWORD=''
+# Use `postgres` as the POSTGRES_HOST for docker, otherwise add ip/domain name
+# POSTGRES_HOST=postgres
+# Same as above, use `redis` as the REDIS_HOST for docker, or ip/domain name
+# REDIS_HOST=redis
+# generate secure password and setup redis accordinngly 
+# REDIS_PASSWORD=''
+# REDIS_AOF_ENABLED=no
+# Use MEDIA_PATH to point to a different location to store the media files
+# MEDIA_PATH=''
 
 # Secrets
-SECRET_KEY='' # generate using 'openssl rand -base64 32'
-SECURITY_PASSWORD_SALT='' # generate using 'openssl rand -base64 32'
+# Generate SECRET_KEY using 'openssl rand -base64 32'
+SECRET_KEY=''
+# Generate SECURITY_PASSWORD_SALT using 'openssl rand -base64 32'
+SECURITY_PASSWORD_SALT=''
 
 # 2FA
-SECURITY_TOTP_SECRETS='' # generate using 'openssl rand -base64 32'
+## Generate SECURITY_TOTP_SECRETS using 'openssl rand -base64 32'
+SECURITY_TOTP_SECRETS='' 
 SECURITY_TWO_FACTOR=True
 SECURITY_TWO_FACTOR_RESCUE_MAIL=''
 SECURITY_TWO_FACTOR_AUTHENTICATOR_VALIDITY=90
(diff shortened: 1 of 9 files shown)


sfpurdy:
I closed my [previous PR](https://github.com/sjacorg/bayanat/pull/19) in favor of this new one.

sjacgit:
Thanks, we'll review.

sjacgit:
We're really sorry about the delay. At the moment we're unable to review and merge this PR as we're considering requiring contributor licence agreement. Please bear with us as we decide on this and we'll get back to you soon.

flask/Dockerfile
```diff
@@ -34,9 +48,9 @@ ARG ROLE
 ENV ROLE=${ROLE}
 RUN echo "Building ${ROLE} container."
 RUN if [ "$ROLE" = "flask" ]; then \
-    apt update -y && apt install -yq python3-dev python3.10-venv postgis; \
+    apt update -y && apt install -yq python3.8-dev python3.8-venv postgis; \
```
zhabiba24, line 51: Do you definitely need postgis in the base image here, isn't that more for postgres rather than python app?

.env-sample
```diff
@@ -59,3 +70,16 @@ DEDUP_TOOL=False
...
+## You should change both of these
+ADMIN_USERNAME=bayanat
+ADMIN_PASSWORD=change_this_password
+
+# Docker Compose Variables
+## Use a different port to reach the services from external locations
+BAYANAT_EXTERNAL_PORT=5000
+NGINX_EXTERNAL_PORT=80
+POSTGRES_EXTERNAL_PORT=5432
+REDIS_EXTERNAL_PORT=6379
```
zhabiba24, line 85: nit: missing a newline here and in a few other spots

sjacgit:
Hello again. 

Again we apologize for the time it took us to get back to you. We'll unfortunately reject this PR. However, some of the issues raised here has already been addressed albeit in a different way.

First of all adding a default username and password is a very bad practice which we've purposely avoided, even at the cost of adding an extra step to the installation. Nonetheless, in the latest version of Bayanat, the first admin user can be created from the browser upon first use. 

The current `docker-compose.yml` has been changed a lot since your changes, although most of your requested changes are not addressed. It's optimized for security, reliability and simplicity but it's also serves as a general template for most users who will use the default settings. You're welcome to create your own to fit your infrastructure, unfortunately we won't incorporate your changes here.  If you feel there are some changes needed to the current compose file which would affect general users please file a bug or new feature request so discuss the issues you're addressing and agree on a solution.

Lastly, regarding the connection dropping issue, could you please file a bug report to so we can better understand this issue? We need to check how and when this issue occur before we're able to merge a fix. This is to ensure there isn't a better solution available from our side. 

Again thank you so much for your interest and contributions.

PR #43, opened 2024-10-07
bugfix: activities options watcher

crnogaj:
Since `update:options` on table already emits (https://vuetifyjs.com/en/api/v-data-table-server/#events-update:options) when options changes we don't need separate watcher for it.
That way it will do just one request instead of two.

### DIFF
--- a/enferno/admin/templates/admin/activity.html
+++ b/enferno/admin/templates/admin/activity.html
@@ -288,10 +288,6 @@
       dialog(val) {
         val || this.close();
       },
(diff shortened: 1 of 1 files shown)


sjacgit:
Thanks for your contribution. We're reviewing this and will get back to you ASAP.

level09 (approved):
Sounds good. we no longer watch the options after migrating to v3

tarekio (approved):

PR #49, opened 2025-01-03
BYNT-1275: CSRF rate limiting

cango91:
## Jira Issue
1. [BYNT-1275](https://syriajustice.atlassian.net/browse/BYNT-1275)

## Description
We were discussing whether to keep databases separate so as not to corrupt production db. Otherwise we'll need to keep track of all keys added during test and remove them after tests are done.

## Checklist
- [X] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/app.py
+++ b/enferno/app.py
@@ -31,7 +31,7 @@ from enferno.admin.models import (
 )
 from enferno.admin.views import admin
 from enferno.data_import.views import imports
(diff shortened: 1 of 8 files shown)


level09 (approved):
All good. it might be better to move lookup proxy headers to settings but we can do it in a later iteration.

tarekio (approved):

PR #50, opened 2025-01-03
Notification system 🚨

cango91:
## Jira Issue
[Phase 1](https://syriajustice.atlassian.net/browse/BYNT-1179)
[Phase 2](https://syriajustice.atlassian.net/browse/BYNT-1180)
[Phase 3](https://syriajustice.atlassian.net/browse/BYNT-1181)
[Phase 4](https://syriajustice.atlassian.net/browse/BYNT-1182)
[Phase 5](https://syriajustice.atlassian.net/browse/BYNT-1183)
[Phase 6](https://syriajustice.atlassian.net/browse/BYNT-1184)
[Phase 7](https://syriajustice.atlassian.net/browse/BYNT-1185)
[Phase 8](https://syriajustice.atlassian.net/browse/BYNT-1186) (PENDING)
[BYNT-1249](https://syriajustice.atlassian.net/browse/BYNT-1249)
[BYNT-1254](https://syriajustice.atlassian.net/browse/BYNT-1254)
## Description
[Brief description of changes]

## Checklist
 - [ ] Tests added/updated
 - [ ] Billing project label added
 - [ ] API Changes (if applicable)
 - [ ] Permissions checked 
 - [ ] Endpoint tests added
## Additional Notes
`GET /admin/api/notifications` return a paginated list of all notifications for the current user (descending by created date). Accepts `page` and `per_page` query parameters. A sample response looks like below
```
{
    "currentPage": 1,
    "hasMore": false,
    "items": [],
    "perPage": 10,
    "total": 0
}
```
`GET admin/api/notifications/unread/count`: return the number of unread notifications for the current user. Will have only one field in JSON response unread_count
`POST admin/api/notifications/<int:id>/read`: mark the notification with id as read, if it belongs to the current user. Will return a message and the notification on success.
`POST admin/api/notifications/send`: A testing endpoint to send an internal test notification to the current user.
`POST admin/api/notifications/send-email`: A testing endpoint to send an email test notification to the current user. Will log error if user doesn't have an email.

### DIFF
--- a/enferno/admin/models/Notification.py
+++ b/enferno/admin/models/Notification.py
@@ -1,0 +1,65 @@
+from datetime import datetime
+from typing import Any
+from enferno.admin.models.utils import check_roles
+from enferno.utils.date_helper import DateHelper
+from enferno.utils.logging_utils import get_logger
+from enferno.extensions import db
+from enferno.utils.base import BaseMixin
+import json
+
+logger = get_logger()
+
+
+class Notification(db.Model, BaseMixin):
+    TYPE_GENERAL = "general"
+    TYPE_UPDATE = "update"
+    TYPE_SECURITY = "security"
+
+    DELIVERY_METHOD_EMAIL = "email"
+    DELIVERY_METHOD_SMS = "sms"
+    DELIVERY_METHOD_INTERNAL = "internal"
+
+    id = db.Column(db.Integer, primary_key=True)
+    user_id = db.Column(db.Integer, db.ForeignKey("user.id"), nullable=False, index=True)
+    user = db.relationship("User", foreign_keys=[user_id], backref="user_notifications")
+    title = db.Column(db.String, nullable=False)
+    message = db.Column(db.Text, nullable=False)
+    notification_type = db.Column(db.String, nullable=False, default=TYPE_GENERAL)
+    read_status = db.Column(db.Boolean, default=False)
+    delivery_method = db.Column(db.String, nullable=False, default=DELIVERY_METHOD_INTERNAL)
+    read_at = db.Column(db.DateTime)
+    is_urgent = db.Column(db.Boolean, default=False)
+
+    __table_args__ = (
+        db.Index("ix_notification_user_read", "user_id", "read_status"),
+        db.Index("ix_notification_user_type", "user_id", "notification_type"),
+    )
+
+    def mark_as_read(self):
+        """Marks the notification as read and sets the read timestamp."""
+        if not self.read_status:
+            self.read_status = True
+            self.read_at = datetime.now()
+            self.save()
+
+    def to_dict(self) -> dict[str, Any]:
+        """Converts notification object to a dictionary representation."""
+        return {
+            "id": self.id,
+            "title": self.title,
+            "message": self.message,
+            "type": self.notification_type,
+            "read_status": self.read_status,
+            "read_at": DateHelper.serialize_datetime(self.read_at) if self.read_at else None,
+            "is_urgent": self.is_urgent,
+            "created_at": DateHelper.serialize_datetime(self.created_at)
+            if self.created_at
+            else None,
+            "updated_at": DateHelper.serialize_datetime(self.updated_at)
+            if self.updated_at
+            else None,
+        }
+
+    def to_json(self) -> str:
+        """Converts notification object to a JSON string."""
+        return json.dumps(self.to_dict())
--- a/enferno/admin/validation/models.py
+++ b/enferno/admin/validation/models.py
@@ -1702,6 +1702,14 @@ class FullConfigValidationModel(ConfigValidationModel):
     ADV_ANALYSIS: bool
     SETUP_COMPLETE: bool = Field(default=True)
     LOCATIONS_INCLUDE_POSTAL_CODE: bool
+    MAIL_ENABLED: bool
+    MAIL_SERVER: Optional[str] = None
+    MAIL_PORT: Optional[int] = None
+    MAIL_USE_TLS: Optional[bool] = None
+    MAIL_USE_SSL: Optional[bool] = None
+    MAIL_USERNAME: Optional[str] = None
+    MAIL_PASSWORD: Optional[str] = None
+    MAIL_DEFAULT_SENDER: Optional[str] = None
     YTDLP_PROXY: Optional[str] = None
     YTDLP_ALLOWED_DOMAINS: list[str] = Field(default_factory=list)
     YTDLP_COOKIES: Optional[str] = None
@@ -1709,6 +1717,17 @@ class FullConfigValidationModel(ConfigValidationModel):
     @model_validator(mode="before")
     def ensure_setup_complete(cls, values):
         values["SETUP_COMPLETE"] = True
+        if values.get("MAIL_ENABLED"):
+            if (
+                not values.get("MAIL_SERVER")
+                or not values.get("MAIL_PORT")
+                or not values.get("MAIL_USERNAME")
+                or not values.get("MAIL_PASSWORD")
+                or not values.get("MAIL_DEFAULT_SENDER")
+            ):
+                raise ValueError(
+                    "MAIL_SERVER, MAIL_PORT, MAIL_USERNAME and MAIL_PASSWORD must be provided if MAIL_ENABLED is True"
+                )
         return values
 
     @field_validator("ITEMS_PER_PAGE_OPTIONS")
--- a/enferno/admin/views.py
+++ b/enferno/admin/views.py
@@ -24,6 +24,8 @@ from flask_security.twofactor import tf_disable
 import shortuuid
 from urllib.parse import urlparse
 
+from enferno.admin.models.Notification import Notification
+from enferno.utils.notification_utils import NotificationUtils
 import enferno.utils.typing as t
 from enferno.admin.models import (
     Bulletin,
@@ -5765,6 +5767,125 @@ def api_logs() -> Response:
             return "Error sending log file", 417
     else:
         return "Log file not found", 404
+
+
+# Notifications
+@admin.route("/api/notifications")
+def api_notifications() -> Response:
+    """
+    Endpoint to return paginated notifications for the current user.
+
+    Returns:
+        - JSON response containing notifications and pagination info
+    """
+    page = request.args.get("page", 1, type=int)
+    per_page = request.args.get("per_page", 10, type=int)
+    status = request.args.get("status")
+
+    # Query notifications for current user, ordered by creation date descending
+    notifications_query = (
+        db.session.query(Notification)
+        .filter(Notification.user_id == current_user.id)
+        .order_by(Notification.created_at.desc())
+    )
+
+    if status:
+        read_status = status.lower() == "read"
+        notifications_query = notifications_query.filter(Notification.read_status == read_status)
+
+    # Paginate the results
+    paginated_notifications = notifications_query.paginate(page=page, per_page=per_page, count=True)
+    unread_count = (
+        db.session.query(Notification)
+        .filter(Notification.user_id == current_user.id, Notification.read_status == False)
+        .count()
+    )
+
+    response = {
+        "items": [notification.to_dict() for notification in paginated_notifications.items],
+        "currentPage": page,
+        "perPage": per_page,
+        "total": paginated_notifications.total,
+        "hasMore": paginated_notifications.has_next,
+        "unreadCount": unread_count,
+    }
+
+    return jsonify(response)
+
+
+# TODO: Remove this endpoint
+@admin.post("/api/notifications/send")
+def api_send_notification() -> Response:
+    """Testing endpoint to send a notification to the current user."""
+    NotificationUtils.send_notification(
+        current_user, "Test Notification", "This is a test notification"
+    )
+    return "Notification sent", 200
+
+
+# TODO: Remove this endpoint
+@admin.post("/api/notifications/send-email")
+def api_send_email_notification() -> Response:
+    """Testing endpoint to send an email notification to the current user."""
+    NotificationUtils.send_notification(
+        current_user,
+        "Test Email Notification",
+        "This is a test email notification",
+        delivery_method="email",
+    )
+    return "Email notification sent", 200
+
+
+@admin.route("/api/notifications/unread/count")
+def api_notifications_unread_count() -> Response:
+    """
+    Endpoint to get the count of unread notifications for the current user.
+
+    Returns:
+        - JSON response containing the count of unread notifications
+        Example: {"unread_count": 5}
+    """
+    unread_count = (
+        db.session.query(Notification)
+        .filter(Notification.user_id == current_user.id, Notification.read_status == False)
+        .count()
+    )
+
+    return jsonify({"unread_count": unread_count})
+
+
+@admin.route("/api/notifications/<int:notification_id>/read", methods=["POST"])
+def api_mark_notification_read(notification_id: int) -> Response:
+    """
+    Endpoint to mark a specific notification as read.
+
+    Args:
+        notification_id: ID of the notification to mark as read
+
+    Returns:
+        - Success response if notification is marked as read
+        - Error response if notification doesn't exist or user doesn't have permission
+    """
+    notification = db.session.get(Notification, notification_id)
+
+    if not notification:
+        return HTTPResponse.NOT_FOUND
+
+    # Verify the notification belongs to the current user
+    if notification.user_id != current_user.id:
+        return HTTPResponse.FORBIDDEN
+
+    try:
+        notification.mark_as_read()
+        return (
+            jsonify(
+                {"message": "Notification marked as read", "notification": notification.to_dict()}
+            ),
+            200,
+        )
+    except Exception as e:
+        logger.error(f"Error marking notification as read: {str(e)}", exc_info=True)
+        return HTTPResponse.INTERNAL_SERVER_ERROR
 
 
 @admin.post("/api/bulletin/web")
--- a/enferno/static/js/components/NotificationsList.js
+++ b/enferno/static/js/components/NotificationsList.js
@@ -1,0 +1,124 @@
+const NotificationsList = Vue.defineComponent({
+    props: [
+        'isInitialLoadingNotifications',
+        'notifications',
+        'hasMoreNotifications',
+        'isLoadingMoreNotifications',
+        'maxTitleLines',
+        'maxSubtitleLines'
+    ],
+    emits: ['readNotification', 'loadNotifications'],
+    data() {
+        return {
+            translations: window.translations
+        };
+    },
+    methods: {
+        getIconFromNotification(notification) {
+            switch (notification?.type?.toLowerCase()) {
+                case 'update':
+                    return 'mdi-update';
+                case 'security':
+                    return 'mdi-security';
+                default:
+                    return 'mdi-bell';
+            }
+        },
+        getDateFromNotification(notification) {
+            return dayjs(notification?.created_at).format('MM/DD/YYYY HH:mm');
+        },
+        getLineClampStyles(lines) {
+            if (!lines) return 'white-space: normal;'
+
+            return `overflow: hidden; display: -webkit-box; -webkit-line-clamp: ${lines}; -webkit-box-orient: vertical; white-space: normal;`
+        }
+    },
+    template: `
+        <div>
+            <v-container v-if="isInitialLoadingNotifications && !notifications" class="text-center">
+                <v-progress-circular color="primary" indeterminate></v-progress-circular>
+            </v-container>
+
+            <v-container v-else-if="!isInitialLoadingNotifications && !notifications?.length">
+                <v-empty-state icon="mdi-bell">
+                    <template v-slot:media>
+                        <v-icon color="surface-variant"></v-icon>
+                    </template>
+
+                    <template v-slot:headline>
+                        <div class="text-h5">
+                            {{ translations.noNotificationsYet_ }}
+                        </div>
+                    </template>
+
+                    <template v-slot:text>
+                        <div class="text-medium-emphasis text-caption">
+                            {{ translations.noNotificationsYetDescription_ }}
+                        </div>
+                    </template>
+                </v-empty-state>
+            </v-container>
+
+            <v-container v-else class="pa-0">
+                <v-card
+                    v-for="(notification, index) in notifications"
+                    :key="index"
+                    density="compact"
+                    :variant="notification?.read_status ? 'flat' : 'tonal'"
+                    :color="notification?.read_status ? '' : 'primary'"
+                    class="mb-2"
+                    :rounded="0"
+                    @click="$emit('readNotification', notification)"
+                >
+                    <v-card-title class="pb-1 d-flex justify-space-between align-center">
+                        <div 
+                            :class="['text-body-1', { 'font-weight-bold': !notification.read_status }]" 
+                            :style="getLineClampStyles(maxTitleLines ?? 1)"
+                            v-html="notification.title"
+                        ></div>
+                        <v-tooltip location="bottom">
+                            <template v-slot:activator="{ props }">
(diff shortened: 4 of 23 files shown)


enferno/admin/validation/models.py
```diff
@@ -1702,13 +1702,32 @@ class FullConfigValidationModel(ConfigValidationModel):
...
+    MAIL_PASSWORD: Optional[str] = None
+    MAIL_DEFAULT_SENDER: Optional[str] = None
     YTDLP_PROXY: Optional[str] = None
     YTDLP_ALLOWED_DOMAINS: list[str] = Field(default_factory=list)
     YTDLP_COOKIES: Optional[str] = None
 
     @model_validator(mode="before")
     def ensure_setup_complete(cls, values):
         values["SETUP_COMPLETE"] = True
+        if values.get("MAIL_ENABLED"):
```
tarekio, line 1720: not sure why this code is here. this validator just ensures SETUP_COMPLETE is always true.

cango91: Because from System Settings we're not sending this field when config changes, but it still needs to exist in our conf file.

enferno/static/js/components/NotificationsList.js
```diff
@@ -0,0 +1,127 @@
...
+    ],
+    emits: ['readNotification', 'loadNotifications'],
+    data() {
+        return {
+            translations: window.translations
+        };
+    },
+    methods: {
+        getIconFromNotification(notification) {
+            switch (notification?.type?.toLowerCase()) {
```
level09, line 19: `notification?.type?` can produce null value, causing unhandled error

level09 (commented):
Great job @apodacaduron , I have added some small comments and suggestions

enferno/static/js/components/NotificationsList.js
```diff
@@ -0,0 +1,127 @@
...
+                case 'update':
+                    return 'mdi-update';
+                case 'security':
+                    return 'mdi-security';
+                default:
+                    return 'mdi-bell';
+            }
+        },
+        getDateFromNotification(notification) {
+            return dayjs(notification?.created_at).format('MM/DD/YYYY HH:mm');
```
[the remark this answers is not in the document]
apodacaduron: Right i added a fallback to return a middle dash `"-"` but if you have a better suggestion for the fallback text let me know

enferno/static/js/components/NotificationsList.js
```diff
@@ -0,0 +1,127 @@
...
+    template: `
+        <div>
+            <v-container v-if="isInitialLoadingNotifications && !notifications" class="text-center">
+                <v-progress-circular color="primary" indeterminate></v-progress-circular>
+            </v-container>
+
+            <v-container v-else-if="!isInitialLoadingNotifications && !notifications?.length">
+                <v-empty-state icon="mdi-bell">
+                    <template v-slot:media>
+                        <v-icon color="surface-variant"></v-icon>
```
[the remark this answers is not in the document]
apodacaduron: The surface-variant is from vuetify but in this case the whole template is not needed so i just removed it, the icon is already set by the line above `<v-empty-state icon="mdi-bell">`

enferno/static/js/mixins/notification-mixin.js
```diff
@@ -0,0 +1,213 @@
...
+  },
+
+  methods: {
+    async markAsReadImportantNotifications() {
+      // TODO: Mark as read important notifications
+      this.isMarkingAsReadImportantNotifications = true;
+      await Promise.all(this.importantNotifications.map(async notification => {
+        await this.readNotification(notification)
+      }))
+      this.isMarkingAsReadImportantNotifications = false;
```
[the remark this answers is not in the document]
apodacaduron: Agreed

enferno/static/js/mixins/notification-mixin.js
```diff
@@ -0,0 +1,213 @@
...
+      page: 1,
+      per_page: 10,
+    },
+    notificationIntervalId: null
+  }),
+
+  async mounted() {
+    this.refetchNotifications();
+
+    this.notificationIntervalId = setInterval(this.refetchNotifications, 60_000);
```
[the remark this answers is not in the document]
apodacaduron: Great idea! thank you this will definitely reduce unneeded api calls

enferno/static/js/components/NotificationsList.js
apodacaduron: Good catch

enferno/static/js/mixins/validations-mixin.js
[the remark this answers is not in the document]
apodacaduron: Oops, forgot to delete it

enferno/templates/notifications_dialog.html
```diff
@@ -0,0 +1,85 @@
...
+                        </v-list-item>
+                    </v-list>
+                    <v-divider v-if="$vuetify.display.mobile" style="z-index: 1;" :vertical="!$vuetify.display.mobile"></v-divider>
+                </div>
+                <v-divider v-if="!$vuetify.display.mobile" style="z-index: 1;" :vertical="!$vuetify.display.mobile"></v-divider>
+            
+                <div class="w-100">
+                    <v-window v-model="currentNotificationTab" style="min-height: calc(100vh - 251px);">
+                        <v-window-item value="all">
+                            <notifications-list
```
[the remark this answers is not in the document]
apodacaduron: Agreed thank you

enferno/static/js/components/NotificationsList.js
```diff
@@ -0,0 +1,127 @@
...
+                    </template>
+
```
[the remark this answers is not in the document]
apodacaduron: !00% sure

enferno/utils/notification_utils.py
```diff
@@ -0,0 +1,230 @@
...
+        # send notification
+        results = []
+        for method in delivery_methods:
+            _, result = NotificationUtils._send_notification(
+                user, title, message, event_config["category"], method, is_urgent
+            )
+            results.append(result)
+            if not result:
+                logger.warning(
+                    f"Failed to send notification for event {event} to user {user.id} using method {method}"
```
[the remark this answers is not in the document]
cango91: @tarekio @level09 do you agree, should we consider some event names sensitive and redact them in logs?

enferno/utils/notification_utils.py
```diff
@@ -0,0 +1,230 @@
...
+from pydantic import ConfigDict
+from enferno.utils.logging_utils import get_logger
+from enferno.utils.email_utils import EmailUtils
+from enferno.admin.constants import Constants
+
+NotificationEvent = Constants.NotificationEvent
+logger = get_logger()
+
+
+class NotificationDeliveryStrategy(ABC):
```
level09, line 16: I think the abstraction/Strategy pattern here is an overkill for our simple use case, instead of defining abstract classes we could simply check a variable i.e delivery_method , also we don't really need to complicate the class with hierarchies 

```
def send_notification(notification: Notification) -> bool:
    """Send notification via appropriate delivery method."""
    if notification.delivery_method == 'email' and notification.user.email:
        return EmailUtils.send_email(
            recipient=notification.user.email,
            subject=notification.title,
            body=f"{notification.title}\n\n{notification.message}",
        )
    elif notification.delivery_method == 'sms':
        # TODO: Implement SMS delivery
        return False
    else:  # internal
        return True

File: enferno/utils/notification_settings.py [hunk 11]
```diff
@@ -0,0 +1,200 @@
+from enferno.admin.constants import Constants
+
+NotificationEvent = Constants.NotificationEvent
+
+
+class NotificationSettings:
```
level09, line 6: I think this class is overcomplicated, with multiple methods for getting, pruning, and enforcing configuration settings. This complexity is unnecessary for the simple goal of loading a config and ensuring secure settings are passed.
currently we have the following methods: 
- `_init_config()`
- `prune_read_only_settings()`
- `_enforce_locked_settings()`
- `get_config()`
- `get_default_config()`
- `get_pruned_default_config()`
- `get_pruned_config()`

Maybe we could simplify it down to something like this : 

```
class NotificationSettings:
    @staticmethod
    def get_config():
        """Get notification configuration with security events always enabled."""
        from enferno.settings import manager
        config = manager.get_config("NOTIFICATIONS")

        # Security events are always enabled and override any user config
        security_events = [
            'LOGIN_NEW_IP', 'PASSWORD_CHANGE', 'TWO_FACTOR_CHANGE',
            'RECOVERY_CODES_CHANGE', 'FORCE_PASSWORD_CHANGE'
        ]

        for event in security_events:
            config[event] = {
                'in_app_enabled': True,
                'email_enabled': True,
                'category': 'security'
            }

        return config

File: enferno/admin/views.py [hunk 12]
```diff
@@ -5762,6 +5912,140 @@ def api_logs() -> Response:
...
+        "total": paginated_notifications.total,
+        "hasMore": paginated_notifications.has_next,
+        "unreadCount": unread_count,
+        "hasUnreadUrgentNotifications": has_unread_urgent_notifications,
+    }
+
+    return jsonify(response)
+
+
+@admin.route("/api/notifications/unread/count")
```
level09, line 5981: why is this endpoint needed ? if the main `/api/notifications` endpoint already provides the `unreadCount`?

we can perhaps delete this and also update the front-end

enferno/static/js/mixins/notification-mixin.js
```diff
@@ -0,0 +1,249 @@
+const notificationMixin = {
```
level09, line 1: This mixin uses over 15 flat data properties to manage state. This makes the component's state difficult to track and understand.

**Current Over-engineered State:**

```javascript
data: () => ({
    isImportantNotificationsDialogVisible: false,
    isInitialLoadingNotifications: false,
    isMarkingAsReadImportantNotifications: false,
    isLoadingMoreNotifications: false,
    isNotificationsDrawerVisible: false,
    isNotificationsDialogFullscreen: false,
    isNotificationsDialogVisible: false,
    unreadNotificationsCount: 0,
    unableToLoadNotifications: false,
    importantNotifications: [],
    notifications: null,
    totalNotifications: 0,
    hasMoreNotifications: false,
    // ... and more
}),
```

**Proposed Simplification:**

```javascript
data() {
  return {
    notifications: {
      items: [],
      importantItems: [],
      unreadCount: 0,
      hasMore: false,
      total: 0,
      // Grouped status flags
      status: {
        loading: false,
        loadingMore: false,
        error: null, // Can store error message string
      }
    },
    ui: {
      drawerVisible: false,
      dialogVisible: false,
      dialogFullscreen: false,
      importantDialogVisible: false,
      currentTab: 'all'
    }
  }
}
```

What do you think?

enferno/static/js/components/NotificationsList.js
```diff
@@ -0,0 +1,162 @@
+const NotificationsList = Vue.defineComponent({
```
level09, line 1: This component accepts too many granular props, making it tightly coupled to the parent's state management implementation.

**Current Over-engineered Props:**

```javascript
props: [
  'isInitialLoadingNotifications',
  'notifications',
  'hasMoreNotifications',
  'isLoadingMoreNotifications',
  'maxTitleLines',
  'maxSubtitleLines',
  'hideScrollFeedback',
  'unableToLoadNotifications'
]
```

**maybe we could simplify like this :**

```javascript
props: {
  notifications: { type: Array, required: true },
  loading: { type: Boolean, default: false },
  loadingMore: { type: Boolean, default: false },
  error: { type: String, default: null },
  hasMore: { type: Boolean, default: false },
  // Group non-essential display options into a config object
  config: {
    type: Object,
    default: () => ({
      maxTitleLines: 1,
      maxSubtitleLines: 2,
      hideScrollFeedback: false
    })
  }
}
```

I think it will make it simpler and less coupled.

enferno/utils/notification_utils.py
```diff
@@ -0,0 +1,230 @@
...
+        # TODO: Implement SMS delivery
+        raise NotImplementedError("SMS delivery is not implemented")
+
+
+class InternalNotificationDeliveryStrategy(NotificationDeliveryStrategy):
+    def send(self, notification: Notification) -> bool:
+        return True
+
+
+class NotificationData(BaseModel):
```
level09, line 46: I think using pydantic here is overkill. lets remove it. internal function calls can be trusted.

enferno/static/js/mixins/notification-mixin.js
```diff
@@ -0,0 +1,249 @@
...
+      page: 1,
+      per_page: 10,
+    },
+    notificationIntervalId: null
+  }),
+
+  async mounted() {
+    this.refetchNotifications();
+
+    // Only poll when document is visible to save resources
```
level09, line 36: This is nice, but polling is still active every 60 sec , we can make polling pause and resume based on page visibility: 


```

// In methods
handleVisibilityChange() {
    if (document.visibilityState === 'visible') {
        this.refetchNotifications(); // Fetch immediately on becoming visible
        this.startPolling();
    } else {
        this.stopPolling();
    }
},
startPolling() {
    if (this.notificationIntervalId) return; // Prevent multiple intervals
    this.notificationIntervalId = setInterval(this.refetchNotifications, 60000);
},
stopPolling() {
    clearInterval(this.notificationIntervalId);
    this.notificationIntervalId = null;
},

// In mounted()
this.refetchNotifications(); // Initial fetch
this.startPolling();
document.addEventListener('visibilitychange', this.handleVisibilityChange);

// In beforeUnmount() (this is less important as we don't really have an actual full SPA
this.stopPolling();
document.removeEventListener('visibilitychange', this.handleVisibilityChange);

**Review by level09 — CHANGES_REQUESTED** [review]

File: enferno/admin/views.py (hunk 12 above)
Comment (cango91) [reply enferno/admin/views.py]: endpoint removed in [a37160b](https://github.com/sjacorg/bayanat/pull/50/commits/a37160bdfc0b0d731435e125c4d7020617ebe93b)

File: enferno/utils/notification_settings.py (hunk 11 above)
Comment (cango91) [reply enferno/utils/notification_settings.py:6]: simplified in [a37160b](https://github.com/sjacorg/bayanat/pull/50/commits/a37160bdfc0b0d731435e125c4d7020617ebe93b).

All prune/enforcement methods removed, NotificationSettings now only serves to override/add-in (first time) security-related notification settings in `settings.py`. It no longer polices`config.json` , any notification settings can be written to config.json, but security-related ones will always be overridden on app startup. 

Removes "in_app_locked" "email_locked" settings as well as "category" from configuration. Category now must be specified when using `send_notification_....` functions. [fix category issues in notification_utils. remove invalid method call…](https://github.com/sjacorg/bayanat/pull/50/commits/4bfbc1a21fd99af895cda7aae8b384262aef5f1c)

File: enferno/utils/notification_utils.py (hunk 10 above)
Comment (cango91) [reply enferno/utils/notification_utils.py]: strategy pattern and abc removed in [a37160b](https://github.com/sjacorg/bayanat/pull/50/commits/a37160bdfc0b0d731435e125c4d7020617ebe93b)

File: enferno/admin/views.py [hunk 17]
```diff
@@ -5762,6 +5912,140 @@ def api_logs() -> Response:
         return "Log file not found", 404
 
 
+# Notifications
+@admin.route("/api/notifications")
```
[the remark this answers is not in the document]
cango91: updated in [a37160b](https://github.com/sjacorg/bayanat/pull/50/commits/a37160bdfc0b0d731435e125c4d7020617ebe93b)

enferno/admin/validation/models.py
level09: Yep the validator should never silently override configuration to True. This code sure should be removed.

enferno/admin/views.py
apodacaduron: Frontend has been adjusted

enferno/static/js/components/NotificationsList.js
apodacaduron: updated!

enferno/utils/config_utils.py
apodacaduron, line 1: I'm gonna remove those settings from frontend as well, will leave a comment when ui is updated

apodacaduron: Updated

level09 (approved):

enferno/tasks/__init__.py
```diff
@@ -122,6 +124,70 @@ def chunk_list(lst: list, n: int) -> Generator[list, Any, None]:
...
+    from enferno.utils.email_utils import EmailUtils
+    from enferno.admin.models.Notification import Notification
+    from datetime import datetime
+
+    # Get the notification record
+    notification = Notification.query.get(notification_id)
+    if not notification:
+        logger.error(f"Notification {notification_id} not found")
+        return False
+
```
level09, line 147: + Also  better guard against duplicate : 

```
if notification.email_sent:
    logger.info(f"Notification {notification_id} already sent; skipping send.")
    return True

File: enferno/tasks/__init__.py [hunk 19]
```diff
@@ -122,6 +124,70 @@ def chunk_list(lst: list, n: int) -> Generator[list, Any, None]:
...
+    except Exception as e:
+        # Handle unexpected errors
+        logger.error(f"Email notification {notification_id} failed: {str(e)}", exc_info=True)
+        success = False
+
+    finally:
+        if success:
+            # record successful email send
+            notification.email_sent = True
+            notification.email_sent_at = datetime.now()
```
level09, line 170: this uses local time, but we use utcnow() across the app, so maybe better unify it as well for consistency ?

enferno/tasks/__init__.py
```diff
@@ -122,6 +124,70 @@ def chunk_list(lst: list, n: int) -> Generator[list, Any, None]:
...
+    notification = Notification.query.get(notification_id)
+    if not notification:
+        logger.error(f"Notification {notification_id} not found")
+        return False
+
+    # Check if user has email (should be validated already, but double-check)
+    if not notification.user.email:
+        logger.error(f"User {notification.user.id} has no email address")
+        return False
+
```
level09, line 152: + Also validate and check Allowed domains here ? why aren't we validating emails against the `MAIL_ALLOWED_DOMAINS` ?

enferno/tasks/__init__.py
```diff
@@ -122,6 +124,70 @@ def chunk_list(lst: list, n: int) -> Generator[list, Any, None]:
...
+        logger.error(f"Email notification {notification_id} failed: {str(e)}", exc_info=True)
+        success = False
+
+    finally:
+        if success:
+            # record successful email send
+            notification.email_sent = True
+            notification.email_sent_at = datetime.now()
+            notification.save()
+            logger.info(
```
level09, line 172: Maybe we should not expose user-identifying emails in logs ? wdyt ?

enferno/tasks/__init__.py
```diff
@@ -122,6 +124,70 @@ def chunk_list(lst: list, n: int) -> Generator[list, Any, None]:
...
+
+    except Exception as e:
+        # Handle unexpected errors
+        logger.error(f"Email notification {notification_id} failed: {str(e)}", exc_info=True)
+        success = False
+
+    finally:
+        if success:
+            # record successful email send
+            notification.email_sent = True
```
level09, line 169: This code is unreachable and will never run. 

success is never defined in this scope.

enferno/migrations/20250718_153735_create_notification_table.sql
```diff
@@ -0,0 +1,35 @@
+-- Migration: Create notification table
+
+-- Create notification table
+CREATE TABLE IF NOT EXISTS notification (
+    id SERIAL PRIMARY KEY,
+    user_id INTEGER NOT NULL,
+    title VARCHAR NOT NULL,
+    message TEXT NOT NULL,
+    category VARCHAR NOT NULL DEFAULT 'general',
```
level09, line 9: doesn't match model's default value `update`

enferno/migrations/20250718_153735_create_notification_table.sql
```diff
@@ -0,0 +1,35 @@
...
+    read_status BOOLEAN DEFAULT FALSE,
+    delivery_method VARCHAR NOT NULL DEFAULT 'internal',
```
level09, line 11: remove

enferno/migrations/20250718_153735_create_notification_table.sql
```diff
@@ -0,0 +1,35 @@
...
+    read_at TIMESTAMP,
+    is_urgent BOOLEAN DEFAULT FALSE,
+    delivery_status VARCHAR NOT NULL DEFAULT 'pending',
```
level09, line 14: remove

enferno/migrations/20250718_153735_create_notification_table.sql
```diff
@@ -0,0 +1,35 @@
...
+    delivery_error TEXT,
```
level09, line 15: remove

enferno/migrations/20250718_153735_create_notification_table.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE INDEX IF NOT EXISTS ix_notification_user_type ON notification (user_id, category);
+CREATE INDEX IF NOT EXISTS ix_notification_user_delivery_status ON notification (user_id, delivery_status);
+
+-- Add foreign key constraint
+ALTER TABLE notification 
+ADD CONSTRAINT fk_notification_user_id 
+FOREIGN KEY (user_id) REFERENCES "user"(id);
+
+-- Rollback script:
+/*
```
level09, line 33: This file is missing columns and has extra non existing columns. and has inconsistent indexes. I suggest revising it based on our latest Notification model.

enferno/admin/models/Notification.py
```diff
@@ -0,0 +1,226 @@
...
+    category = db.Column(db.String, nullable=False, default=NotificationCategories.UPDATE.value)
+    read_status = db.Column(db.Boolean, default=False)
+    read_at = db.Column(db.DateTime)
+    is_urgent = db.Column(db.Boolean, default=False)
+
+    # Email delivery tracking
+    email_enabled = db.Column(db.Boolean, default=False)
+    email_sent = db.Column(db.Boolean, default=False)
+    email_sent_at = db.Column(db.DateTime)
+
```
level09, line 36: I noticed you removed email_error etc .. is it intentional ? we dont need to track email sending error in the db ? or pefer to only log them ?

level09 (changes requested):

enferno/tasks/__init__.py
```diff
@@ -122,6 +124,70 @@ def chunk_list(lst: list, n: int) -> Generator[list, Any, None]:
...
+
+    Returns:
+        bool: True if email was sent successfully, False otherwise
+    """
+    from enferno.utils.email_utils import EmailUtils
+    from enferno.admin.models.Notification import Notification
+    from datetime import datetime
+
+    # Get the notification record
+    notification = Notification.query.get(notification_id)
```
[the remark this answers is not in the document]
tarekio: Addressed.

enferno/admin/models/Notification.py
tarekio: Only log is enough. We can tell that there's an error if email_enabled and not email_sent. We can also find the logs easily as we're logging notif id.

enferno/tasks/__init__.py
tarekio: Indeed that's a valid scenario. I don't think we should not send an email from there, but we should handle this in a future PR. We need to make some tweaks anyway regarding emails, like allowing users to set their own for example, and we can address this at the same time. However, I don't think by adding validation here. We will have to think of a proper solution (invalidating all emails, force users to change emails, etc...).

enferno/tasks/__init__.py
tarekio: Indeed, fixed.

enferno/tasks/__init__.py
tarekio: Removed the whole try except as it's not needed in the first place.

level09 (approved):
This is good, Everything is fixed.

PR #58, opened 2025-01-14
System administration new design

apodacaduron:
## Jira Issue
1. [1333](https://syriajustice.atlassian.net/browse/BYNT-1333)

## Description
image

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes


enferno/static/js/common/config.js
```diff
@@ -8,22 +8,32 @@ const validationRules = {
 // global vuetify config object passed to most pages of the system
 const vuetifyConfig = {
     defaults: {
-        VTextField: {
```
level09, line 11: This seems to affect elements globally 
image

this causes inconsistency in the design, compared to the previous one below (this screenshot from new bulletin dialog) 

image

enferno/admin/templates/admin/system-administration.html
```diff
@@ -888,43 +926,44 @@ <h4 class="mt-3 subtitle-2 mb-2">{{ _('Full Location Generation Settings') }}</h
 
                 },
                 generateConfigDiff(delta) {
```
level09, line 928: image

Diff shows a list opening `[` without closing one. can we add that or remove opening square bracket

level09 (commented):
A few small notes: 

1.  Can we make sure that global vuetify styles are not causing inconsistencies across the design (unless we plan to update everything later) 

2. Could we also make the confirm button using primary color style (elevated perhaps) 
image

3. I noticed that if you visit configurations and simply pass by the maps tab, the save button activates automatically, can we fix this ? 

4. when we save , I don't like the overlay text design as it is not readable 

image

can we improve this ? maybe add a solid container for the text

enferno/static/js/common/config.js
apodacaduron: Feedback addressed

enferno/admin/templates/admin/system-administration.html
apodacaduron: Feedback addressed

level09 (commented):
This looks great. but can we fix this scenario in config state : 


1. go to system settings , `save settings` button is disabled 
2. go to maps tab, -> `save settings` button gets activated. 

I think our geomap component needs a fix as it emits or changes the state of our config.

enferno/admin/templates/admin/component-data.html
```diff
@@ -1,512 +1,506 @@
...
+                    violation: {
+                        text: "{{ _('Violation Categories') }}",
+                        nestedTabsMap: {
+                            ['potential-violation']: { text: "{{ _('Potential Violation') }}" },
+                            ['claimed-violation']: { text: "{{ _('Claimed Violation') }}" }
+                        }
+                    },
+                },
+
                 alHeaders: [
```
level09, line 329: **Can we replace Repetitive Data Structures with Factory Functions**

### ❌ **Before: Repetitive Arrays/Objects**

```javascript
// 50+ lines of nearly identical configurations
userHeaders: [
    { title: 'ID', value: 'id' },
    { title: 'Name', value: 'name' },
    { title: 'Email', value: 'email' },
    { title: 'Actions', value: 'actions', align: 'end' }
],

productHeaders: [
    { title: 'ID', value: 'id' },
    { title: 'Name', value: 'name' },
    { title: 'Price', value: 'price' },
    { title: 'Actions', value: 'actions', align: 'end' }
],

orderHeaders: [
    { title: 'ID', value: 'id' },
    { title: 'Customer', value: 'customer' },
    { title: 'Total', value: 'total' },
    { title: 'Actions', value: 'actions', align: 'end' }
],
// ... 10+ more identical structures
```

### **After: Single Factory Function**

```javascript
// 8 lines replaces 50+ lines
createHeaders(type, customFields = []) {
    const base = [{ title: 'ID', value: 'id' }];
    
    const fieldMaps = {
        user: [{ title: 'Name', value: 'name' }, { title: 'Email', value: 'email' }],
        product: [{ title: 'Name', value: 'name' }, { title: 'Price', value: 'price' }],
        order: [{ title: 'Customer', value: 'customer' }, { title: 'Total', value: 'total' }]
    };
    
    return [...base, ...(fieldMaps[type] || customFields), { title: 'Actions', value: 'actions', align: 'end' }];
}

// Usage: createHeaders('user'), createHeaders('product'), etc.
```

**Result**: **90% reduction** in header definitions

level09 (commented):
## System Administration

Small issue I Found, If i go to System Administration Dashboard, Maps tab and change only coordinates, the diff shows Diameter change to 1000 , maybe we need to fix default value bindings ? let me know if this is better done on the backend API. 


## Tab URLs and Router 

It is nice to have the URL respond to the current active tab, but the history is not really functional, can we make it work ? like we do with drawers history ? (perhaps replace `replace` with `push`)

tarekio (changes requested):
Good overall, Import tools tab needs some work to match design.

apodacaduron:
@tarekio updated, import tools should better match figma now

PR #65, opened 2025-01-20
bug: BYNT-1332 - Web Import Fails for Content With Long Titles

cango91:
## Jira Issue
1. [BYNT-1332](https://syriajustice.atlassian.net/browse/BYNT-1332)

## Description
Web Import changes: 
- Truncates long video titles to 255 characters when creating bulletin and (if truncation happened) adds the non-truncated original to the comments to make it available for searches.

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated (if needed)
- [ ] New strings prepared for translations

## API Changes (if applicable)
- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes
[Any other relevant information]

[BYNT-1332]: https://syriajustice.atlassian.net/browse/BYNT-1332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #67, opened 2025-01-20
Bynt-1328/1332 Preserving Arabic titles media import/Web Import Can't Import Content With Long Titles

cango91:
## Jira Issue
1. [BYNT-1328](https://syriajustice.atlassian.net/browse/BYNT-1328)
2. [BYNT-1332](https://syriajustice.atlassian.net/browse/BYNT-1332)

## Description
Changes:
- **Sheet Import:** CSV upload endpoint now follows the same approach to unicode characters in uploaded filename as media upload endpoint.
- **Upload Endpoints:** `/admin/api/media/chunk` and `/import/api/csv/upload` now include `original_filename` in their response.
- **Media Import:** for `mode==1` (Upload mode), use `original_filename` to derive Bulletin/Media titles. Truncate long titles (>255) and add original title to description if truncation happened.

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

[BYNT-1328]: https://syriajustice.atlassian.net/browse/BYNT-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[BYNT-1332]: https://syriajustice.atlassian.net/browse/BYNT-1332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


enferno/data_import/utils/media_import.py
```diff
@@ -667,6 +670,16 @@ def create_bulletin(self, info: dict) -> None:
...
+        if info.get("original_filename") and info.get("filename") != info.get("original_filename"):
+            if bulletin.description:
+                bulletin.description += (
+                    f"<br /><strong>Original filename:</strong> {info.get('original_filename')}"
+                )
+            else:
+                bulletin.description = (
+                    f"<strong>Original filename:</strong> {info.get('original_filename')}"
+                )
+
```
tarekio, line 682: This logic here is wrong. The original file name should go to the title, unless it's too long, in which case we cut it for the title and add the full name to the description.

tarekio: also it seems this logic is being repeated a lot (if += else =), let's refactor into a method.

cango91: description update extracted into function.

cango91: The logic here was so that the existing pipeline would not be changed. Since we're now modifying how we derive the title, the introduction of `original_filename` might be unwarranted. Instead, we could keep the `name` field (we're currently setting it to the uploaded file's server-side filename, originally dropzone sets this field to the file's client-side filename) unmodified on the frontend because the backend already used it to derive title in Upload mode.

cango91:
Merged changes from [PR: 65](https://github.com/sjacorg/bayanat/pull/65) closing it.

PR #68, opened 2025-01-20
BYNT-1351: Gracefully handle missing Whisper requirements

cango91:
## Jira Issue
1. [BYNT-1351](https://syriajustice.atlassian.net/browse/BYNT-1351)

## Description
Update `/import/api/whisper/languages` to gracefully handle if `openai-whisper` package is missing

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/data_import/views.py
+++ b/enferno/data_import/views.py
@@ -466,6 +466,10 @@ def api_whisper_models() -> Response:
 @roles_required("Admin")
 def api_whisper_languages() -> Response:
     """Returns the list of whisper languages."""
-    from whisper.tokenizer import TO_LANGUAGE_CODE
+    try:
+        from whisper.tokenizer import TO_LANGUAGE_CODE
 
-    return jsonify({"languages": TO_LANGUAGE_CODE})
+        return jsonify({"languages": TO_LANGUAGE_CODE})
+    except Exception as e:
+        logger.error(e, exc_info=True)
+        return "Error getting whisper languages", 417


cango91:
Updated the mixin to respect the `window.__TRANSCRIPTION_ENABLED__` set by the template, so it won't send request when transcription is disabled.
The API endpoint `/import/api/whisper/languages` , if queried nonetheless, will respond with `417` if either `TRANSCRIPTION_ENABLED` config is `False` or there is an error importing `whisper` and log the error.

enferno/data_import/views.py
```diff
@@ -466,6 +466,13 @@ def api_whisper_models() -> Response:
 @roles_required("Admin")
 def api_whisper_languages() -> Response:
     """Returns the list of whisper languages."""
-    from whisper.tokenizer import TO_LANGUAGE_CODE
+    if not current_app.config.get("TRANSCRIPTION_ENABLED"):
+        return "Transcription is not enabled.", 417
 
-    return jsonify({"languages": TO_LANGUAGE_CODE})
+    try:
+        from whisper.tokenizer import TO_LANGUAGE_CODE
```
tarekio, line 473: Th line will fail if the package is missing but the transcription is enabled.

tarekio (changes requested):
This doesn't deal with the source of the issue which is the missing package. 

We need to disable this setting in sys settings if package is not installed and tell the user about the issue.

PR #70, opened 2025-01-20
Add title max length warning

apodacaduron:
## Jira Issue
1. [1300](https://syriajustice.atlassian.net/browse/BYNT-1300)

## Description
Add max length to some fields like Title based on some of the backend models.
Centralize common validations in config.js and refactor existing validation rules

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes


tarekio (approved):

level09 (commented):
This is nice. but there I think there is a little discrepancy in the validation logic, 

**Previously:**
In the create bulletin dialog, if you filled only the translation in the "Original Title" field and moved to the next field, the original field would immediately be marked as invalid (red), providing instant feedback that at least one of the fields needs to be filled.

**With the new validation logic:**
If you fill only the translation and move to the next field, the original field is not marked as invalid (no red border or error) until you actually try to submit the form, still it won't mark the required field as red but only shows a snackbar.
This delays feedback to the user and could lead to confusion, 

To improve clarity, maintainability, and ux, I recommend introducing a dedicated validation rule, such as “at least one is required”—for bilingual fields. This would make the intent explicit in the code and ensure users receive immediate, consistent feedback as soon as they leave the field.

apodacaduron:
@level09 Great find, thanks for the suggestion! I added the new rule `atLeastOneRequired` and explicitly pass a value to ensure that at least one of the fields is always required.

level09 (approved):
👍

PR #75, opened 2025-01-22
Duplicated rows when creating records

apodacaduron:
## Jira Issue
1. [1353](https://syriajustice.atlassian.net/browse/BYNT-1353)

## Description
**Problem:**
Currently, when creating records on pages such as bulletins, actors, etc., the newly created item is pushed directly into the table without an ID. This causes occasional visual artifacts where the table displays duplicate rows.

**Solution:**
This PR resolves the issue by using the server response, which includes the correct ID, to update the table. This ensures the integrity of the data displayed and eliminates duplicate rows caused by the missing ID.

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/admin/templates/admin/actors.html
+++ b/enferno/admin/templates/admin/actors.html
@@ -1582,8 +1582,8 @@
                         axios.post("/admin/api/actor/", {
                             item: this.editedItem
                         }).then(response => {
-                            this.items.push(this.editedItem);
-                            this.showSnack(response.data);
+                            this.items.push(response.data.data);
+                            this.showSnack(response.data.message);
                             this.refresh();
                             this.close();
                         }).catch(err => {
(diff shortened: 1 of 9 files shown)


tarekio (commented):
You didn't cover users.

tarekio (changes requested):
This PR creates a mismatch between the api responses in create and update.

apodacaduron:
@tarekio I’ve reverted the backend changes and removed the code that was manually adding the new item to the table. Since the table data is already being refreshed after creating a new record, this should resolve the issue with duplicate entries.

tarekio (approved):
Works well.

tarekio (approved):
Perfect. We can address other endpoints in a future PR if needed.

PR #110, opened 2025-06-13
Search actor tags

cango91:
## Jira Issue
1. [BYNT-1362](https://syriajustice.atlassian.net/browse/BYNT-1362)

## Description
+ Adds `tags` as an optional valid search parameter for Actors, where search capabilities are modeled after the Bulletin tags search.
+ Minimal frontend changes to add tags to actor search
+ API documentation updated

## Checklist
- [ ] Tests added/updated
- [X] Documentation updated (if needed)
- [X] New strings prepared for translations

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/static/js/components/ActorSearchBox.js
+++ b/enferno/static/js/components/ActorSearchBox.js
@@ -86,8 +86,45 @@ const ActorSearchBox = Vue.defineComponent({
                   :label="translations.notContains_"
                   clearable
               ></v-text-field>
+              
+              <div class="d-flex align-center">
+                <v-combobox
+                    v-model="q.tags"
+                    :label="translations.inTags_"
+                    multiple
+                    chips
+                    closable-chips
+                    small-chips
+                    clearable
+                ></v-combobox>
+
+                <v-checkbox :label="translations.any_" dense v-model="q.opTags" color="primary" small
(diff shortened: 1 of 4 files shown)


enferno/static/js/components/ActorSearchBox.js
apodacaduron, line 1: Hi! 👋 We should consider removing deprecated props like `small`, `dense`, and `small-chips`. These no longer have any effect in Vuetify 3, so keeping them around might cause confusion or imply functionality that doesn't exist anymore.

level09: 💯 we better remove them

cango91: Removed all `small` `small-chips` and `dense` props 👍

tarekio (approved):
Works well, need to confirm with user testing though.

level09 (commented):
@cango91  for actor tags, data is likely will be big. we will need to add a proper index for actor tags

I suggest GIN index with array_ops operator class , that is good for array operations. 

Let's add :
1. sqlalchemy index 
2. migration script for addding the index

lets make sure the index name is the same in both cases.

level09 (approved):

PR #114, opened 2025-06-18
Implement better handling of server responses in handle request error

apodacaduron:
## Jira Issue
1. [BYNT-1366](https://syriajustice.atlassian.net/browse/BYNT-1366)

## Description
Currently snackbar will render the html response as is, this PR better detects html response and replaces it with a plain string "An error ocurred"

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

[BYNT-1366]: https://syriajustice.atlassian.net/browse/BYNT-1366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


enferno/static/js/common/config.js
```diff
@@ -153,29 +153,36 @@ axios.interceptors.response.use(
...
         let message = '';
-        for(const field in errors){
-            let fieldName = field;
-            if (fieldName.startsWith('item.')){
-                fieldName = fieldName.substring(5);
-            }
-            message += `<strong style="color:#b71c1c;">[${!fieldName.includes("__root__") ? fieldName : 'Validation Error'}]:</strong> ${errors[field]}<br/>`;
+        for (const field in errors) {
+            let fieldName = field.startsWith('item.') ? field.substring(5) : field;
+            const safeText = `${errors[field]}`.replace(/<\/?[^>]+(>|$)/g, ''); // strip all HTML
```
level09, line 167: I think Regexp can be problematic with parsing html, it is not reliable and might be a performance bottleneck. 

Let's instead do do this : 

1. **Check the Content-Type header**  

   ```js
   const ct = error.response?.headers['content-type'] || '';
   if (ct.includes('text/html')) {
     return getInfraMessage(error.response.status);
   }
   ```

2. **Fallback to DOMParser, much better than regex**  

   ```js
   const text = await error.response.text();
   const doc = new DOMParser().parseFromString(text, 'text/html');
   if (doc.body.children.length) {
     return getInfraMessage(error.response.status);
   }
   ```

### 


Then , we don't have to (should not) deliver the infrastructure error to the user so no need to parse the messages from html. instead, let's implement status based messages like this : 

```js
 switch (response.status) {
    case 502:
    case 503:
      return 'Service temporarily unavailable, please try again shortly.';
    case 500:
      return 'Server error, please try again or contact support.';
    default:
      return 'Unexpected error, please contact support if the issue persists.';
  }
```

and most importantly, let's tell the server we only accept JSON, reducing the chance of HTML fallbacks:

```js
axios.defaults.headers.common['Accept'] = 'application/json';
```

level09 (changes requested):

level09 (approved):

PR #136, opened 2025-07-15
Bynt 1386 add password validation checks back to UI

apodacaduron:
## Jira Issue
1. [BYNT-1386](https://syriajustice.atlassian.net/browse/BYNT-1386)

## Description
Add password validations checks to change_password.html and users.html

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/static/js/common/config.js
+++ b/enferno/static/js/common/config.js
@@ -19,6 +19,29 @@ const validationRules = {
         const defaultMessage = window.translations.pleaseEnterAValidNumber_;
         return v => !v || /^\d+$/.test(v) || message || defaultMessage;
     },
+    matchesField: (otherValue, message) => {
+        const defaultMessage = window.translations.fieldsDoNotMatch_;
+        return v => v === otherValue || message || defaultMessage;
+    },
+    checkPassword: (message) => {
+        const defaultMessage = window.translations.passwordTooWeak_;
+        let timeout;
+      
+        return (v) => {
+          return new Promise((resolve) => {
+            clearTimeout(timeout);
+      
+            timeout = setTimeout(async () => {
+              try {
+                await axios.post('/admin/api/password/', { password: v });
+                resolve(true);
+              } catch (err) {
+                resolve(message || defaultMessage);
+              }
+            }, 350);
+          });
+        };
+    }
 };
 
 // Helper functions
--- a/enferno/templates/security/change_password.html
+++ b/enferno/templates/security/change_password.html
@@ -1,69 +1,79 @@
 {% extends 'layout.html' %}
 {% block content %}
-    <v-main>
-        <v-container fluid>
+<v-main>
+    <v-container fluid>
+        <v-card>
+            <v-toolbar color="primary" title="{{ _fsdomain('Change password') }}"></v-toolbar>
 
-            <v-card>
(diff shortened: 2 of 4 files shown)


tarekio (approved):
Good.

enferno/static/js/common/config.js
```diff
@@ -19,6 +19,31 @@ const validationRules = {
...
+    },
+    checkPassword: ({ onResponse }) => {
+        const defaultMessage = window.translations.passwordTooWeak_;
+        
+      
+        return (v) => {
+          return new Promise((resolve) => {
+            clearTimeout(passwordCheckTimeout);
+      
+            passwordCheckTimeout = setTimeout(async () => {
```
level09, line 34: I feel there is a small design bug in password strength check logic here,  the isStrongPassword expects a boolean but we pass an object in case of error, due to `===` , this still works fine not because of boolean check but because of type check failure 😆, 

I think it is cleaner to pass the correct boolean though. 

```
    onResponse(false);
```

enferno/templates/security/change_password.html
```diff
@@ -1,69 +1,87 @@
...
+                        :rules="[validationRules.matchesField(password)]"></v-text-field>
+                </v-card-text>
 
-                    </v-card-text>
-                </form>
-            </v-card>
-        </v-container>
-    </v-main>
+                <v-card-actions class="pa-4 justify-end">
+                    <v-btn type="submit" variant="elevated" color="primary">
```
level09, line 38: Small suggestion: 

I feel we are doing unnecessary loop here, using submit button and submit handler on the form. then blocking submission to validate. instead, 

let's remove the submit handler and use a normal button

`<v-btn @click="validateAndSave" color="primary" variant="elevated">Save</v-btn>`

then 
```
async validateAndSave() {
    const { valid, errors } = await this.$refs.form.validate();
    if (valid) {
        this.save();
    } else {
        this.showSnack("Please review the form for errors.");
        scrollToFirstError(errors);
    }
}
```

I think this is more vue friendly and simpler.

enferno/templates/security/change_password.html
```diff
@@ -1,69 +1,87 @@
...
+</v-main>
 {% endblock content %}
 
+
 {% block js %}
-    <script>
-        const {createApp} = Vue;
-        const {createVuetify} = Vuetify;
+<script>
+    const { createApp, ref } = Vue;
```
level09, line 63: don't think we need/use ref anywhere

level09 (commented):
This PR is good

enferno/templates/security/change_password.html
apodacaduron: Right it was imported by mistake, fixed

enferno/templates/security/change_password.html
apodacaduron: Updated

enferno/static/js/common/config.js
apodacaduron: Updated

level09 (approved):
The Button is still using type=submit but we can improve this later

apodacaduron:
@level09 i did a refactor on `change_password.html` to include the `@click` and the `validateAndSave` function

PR #143, opened 2025-07-21
Make event title optional

apodacaduron:
## Jira Issue
1. [BYNT-1397](https://syriajustice.atlassian.net/browse/BYNT-1397)

## Description
Events can’t be saved without providing a title, we don’t need always to add a title for the event.

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/static/js/components/EventsSection.js
+++ b/enferno/static/js/components/EventsSection.js
@@ -160,7 +160,6 @@ const EventsSection = Vue.defineComponent({
                                             :label-original="translations.title_"
                                             :label-translation="translations.titleAr_"
                                             :rules="[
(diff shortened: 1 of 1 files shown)


level09:
@tarekio This will result in a strange outcome. users can add empty events with no data. 

Is that okay or desired ?

level09 (approved):
This is technically correct. 

But it will produce bad UX , blank events would be confusing.

tarekio:
> @tarekio This will result in a strange outcome. users can add empty events with no data.
> 
> Is that okay or desired ?

Agree, but title is not required. Let's address this in another PR, we'll add better validation for event.

PR #150, opened 2025-08-05
Filter event types in map visualization

apodacaduron:
## Jira Issue
1. [BYNT-1393](https://syriajustice.atlassian.net/browse/BYNT-1393)

## Description
Add a menu with all actor events. By default, let’s show all event types. We can allow the user to exclude event types as needed, and when they are removed, the visualization should be updated to remove events with those types from view.

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/static/js/components/GlobalMap.js
+++ b/enferno/static/js/components/GlobalMap.js
@@ -7,6 +7,14 @@ const GlobalMap = Vue.defineComponent({
       default: true,
     },
   },
+  computed: {
+    uniqueEventTypes() {
+      return [...new Set(this.locations.map(loc => loc.eventtype).filter(Boolean))];
+    },
+    filteredLocations() {
+      return this.locations.filter(loc => this.selectedLocations.includes(loc.eventtype));
+    }
+  },
 
   data: function () {
     return {
@@ -27,6 +35,7 @@ const GlobalMap = Vue.defineComponent({
       attribution: '&copy; <a target="_blank" href="http://osm.org/copyright">OpenStreetMap</a> contributors',
       googleAttribution: '&copy; <a href="https://www.google.com/maps">Google Maps</a>, Imagery ©2025 Google, Maxar Technologies',
       measureControls: null,
+      selectedLocations: [],
     };
   },
 
@@ -41,10 +50,15 @@ const GlobalMap = Vue.defineComponent({
       if (val?.length || val !== old) {
         this.locations = val;
         this.fitMarkers();
+        this.selectedLocations = [...this.uniqueEventTypes];
       }
       if (val.length === 0) {
         this.map.setView([this.lat, this.lng]);
       }
+    },
+
+    selectedLocations(val) {
+      this.fitMarkers();
     },
 
     locations() {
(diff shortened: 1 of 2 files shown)


apodacaduron:
@tarekio button repositioned and tooltip added 😅

level09 (changes requested):
There is a bug in Popup lifecycle management, scenario below, 


https://github.com/user-attachments/assets/93ab148a-2fbd-4d0c-883a-6ac517a33ab8


I think this is what's happening : 

1. **Initial State**: Markers are created with popups bound to them using `marker.bindPopup()`
2. **User Interaction**: User opens a popup by clicking on a marker
3. **Filter Applied**: User changes event type filter via the new dropdown menu
4. **Problem**: The `selectedLocations` watcher triggers `fitMarkers()`, which calls `this.markerGroup.clearLayers()` (line 151)
5. **Critical Issue**: This removes all markers from the map, **including the marker that has an open popup**
6. **Leaflet State Corruption**: The popup remains open but its parent marker no longer exists on the map
7. **Animation Trigger**: When user double-clicks or zooms, Leaflet tries to animate the popup position
8. **Error**: Popup animation code tries to access `_latLngToNewLayerPoint()` on a null map reference because the popup's parent marker was removed

enferno/static/js/components/GlobalMap.js
```diff
@@ -154,59 +170,63 @@ const GlobalMap = Vue.defineComponent({
...
+        if (this.map.hasLayer && this.map.hasLayer(this.markerGroup)) {
+          this.map.removeLayer(this.markerGroup);
         }
+        this.lastBounds = null;
       }
 
-      this.map.invalidateSize();
+      // Only invalidate once (needed if container changed); can be removed if not needed
+      // small timeout helps Leaflet recalc right after DOM changes
+      setTimeout(() => this.map.invalidateSize(), 50);
```
level09, line 229: Do we need to invalidate the size on every filter change @tarekio I think it is not great UX for this use case.

enferno/static/js/components/GlobalMap.js
```diff
@@ -154,59 +170,63 @@ const GlobalMap = Vue.defineComponent({
...
           this.addEventRouteLinks(eventLocations);
         }
 
+        // Ensure markerGroup is added to the map
+        if (!this.map.hasLayer(this.markerGroup)) {
+          this.markerGroup.addTo(this.map);
+        }
+
+        // Fit bounds only if it changed (avoid refit flicker)
+        const bounds = this.markerGroup.getBounds();
```
level09, line 191: I think we can keep bounds fixed when filtering events. @tarekio what do you think ?

apodacaduron:
> There is a bug in Popup lifecycle management, scenario below,
> 
>  CleanShot.2025-08-09.at.19.37.24.mp4 
> I think this is what's happening :
> 
> 1. **Initial State**: Markers are created with popups bound to them using `marker.bindPopup()`
> 2. **User Interaction**: User opens a popup by clicking on a marker
> 3. **Filter Applied**: User changes event type filter via the new dropdown menu
> 4. **Problem**: The `selectedLocations` watcher triggers `fitMarkers()`, which calls `this.markerGroup.clearLayers()` (line 151)
> 5. **Critical Issue**: This removes all markers from the map, **including the marker that has an open popup**
> 6. **Leaflet State Corruption**: The popup remains open but its parent marker no longer exists on the map
> 7. **Animation Trigger**: When user double-clicks or zooms, Leaflet tries to animate the popup position
> 8. **Error**: Popup animation code tries to access `_latLngToNewLayerPoint()` on a null map reference because the popup's parent marker was removed
> 
> Another UX issue is that when filtering is applied, markers are recreated and their colors change which is confusing.

@level09 i reverted some changes and it should work better now

enferno/static/js/components/GlobalMap.js
```diff
@@ -110,34 +123,44 @@ const GlobalMap = Vue.defineComponent({
...
         let eventLocations = [];
 
-        const locationsWithCoordinates = this.locations.filter(loc => loc.lat && loc.lng);
+        const locationsWithCoordinates = visibleLocations.filter(loc => loc.lat && loc.lng);
 
         for (const loc of locationsWithCoordinates) {
 
-          let mainStr = false;
           if (loc.main) {
             mainStr = this.translations.mainIncident_;
```
level09, line 165: is this used anywhere ?

enferno/static/js/components/GlobalMap.js
```diff
@@ -110,34 +123,44 @@ const GlobalMap = Vue.defineComponent({
...
-    redraw() {
-      this.map.invalidateSize();
+      if (this.map.getZoom() > 14) {
+        // flyout of center when map is zoomed in too much (single marker or many dense markers)
+        this.map.flyTo(this.map.getCenter(), 10, { duration: 1 });
+      }
     },
+    clearAllLayers() {
+      // Remove marker cluster group
+      this.markerGroup.clearLayers();
```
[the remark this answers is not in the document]
apodacaduron: Updated should not give an error anymore on empty map

enferno/static/js/components/GlobalMap.js
apodacaduron: mainStr is not used, it has been deleted

PR #180, opened 2025-09-07
Make session insert idempotent via Postgres ON CONFLICT

level09:
### Overview 

  - Problem: UniqueViolation on sessions.session_token when user_authenticated fires multiple times or concurrently.
  - Fix: Use Postgres upsert (ON CONFLICT ON session_token DO UPDATE) to update last_active, ip, meta, is_active instead of inserting a duplicate.

### Testing 

  - Login, open multiple tabs/refresh rapidly; no DB errors.
  - Admin “User Sessions” view still shows/updates the active session.

### DIFF
--- a/enferno/user/views.py
+++ b/enferno/user/views.py
@@ -14,6 +14,9 @@ from enferno.admin.constants import Constants
 from enferno.settings import Config
 from enferno.user.forms import ExtendedLoginForm
 from enferno.user.models import User, Session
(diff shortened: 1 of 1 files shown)


tarekio:
Closing as it's not an ideal solution with the current approach (no dup sessions allowed). #181 should address the issue at root, we can revisit in the future if we detect issues.

PR #191, opened 2025-09-13
fix: adjust search request data structure for specific endpoint

tarekio:
## Jira Issue
1. https://syriajustice.atlassian.net/browse/BYNT-1461

## Description
Amend q in requestData per search endpoint. This is to accomodate the incident endpoint which accepts one query (vs A/B which accept multiple queries with bool ops).

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes


tarekio:
> Looks good, this fixes the immediate issue cleanly. in the future we could improve this by adding a `queryFormat` property to each components instead of hardcoded string matching with endpoints, I think it would be less fragile and offer better separation.

Done :saluting_face:

level09 (approved):

PR #193, opened 2025-09-18
Add scroll to menu in system administration page

apodacaduron:
## Jira Issue
1. [BYNT-1467](https://syriajustice.atlassian.net/browse/BYNT-1467)

#### Issue
On the **System Administration** page, the menu becomes partially inaccessible on low-resolution screens. During usability testing, some users didn’t notice the notification at all. Additionally, other panels can overflow on smaller screens, making some content hard or impossible to access.

#### Solution
- Enabled scrolling on the System Administration menu.  
- Added buttons inside the menu to scroll directly to the top or bottom.  
- Added scrollbars to other panels on smaller screens to ensure all content is accessible.

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

[BYNT-1467]: https://syriajustice.atlassian.net/browse/BYNT-1467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tarekio (changes requested):
Let's revert to normal scroll. Also, the menu seems to be squeezed, let's restore previous look, and use primary color for selection.

tarekio (approved):
This is ok for now, we can improve more in the future.

enferno/admin/templates/admin/system-administration.html
```diff
@@ -1449,9 +1449,9 @@ <h4 class="subtitle-2 mb-n1">{{ _('Enable Mail System') }}</h4>
                 }
             }
         });
-        app.component('geo-map', GeoMap);
-        app.component('show-details', ShowDetails);
-        app.component('diff-renderer', DiffRenderer);
+        app.component('GeoMap', GeoMap);
```
level09, line 1452: I think those are unnecessary / related but we have to unify all our components registration across code at some point. we have mixed them everywhere 🙃

level09 (approved):
I think this is a good simple fix for the issue. but I think the UX is still not ideal, not because of CSS but because of MAC defaults (hides scrollbars) so there is no way for the user to tell if he can scroll of not. 


Maybe we can implement a simple UI enhancement here ? like show a little tiny UI detail when there is a scrollbar? 

https://github.com/user-attachments/assets/fee63344-c7f4-45f9-975f-768c805f992f

PR #199, opened 2025-09-27
Automatic Updates for Bayanat

level09:
## Summary
- Implements comprehensive automatic update system for Bayanat
- Adds version management and update checking capabilities
- Integrates update functionality into the admin interface

## Key Features
- Automatic version checking against GitHub releases
- One-click update process from the admin panel
- Rollback capability for failed updates
- Comprehensive logging and error handling
- Security considerations with signature verification support

## Implementation Details
- New update module with version comparison logic
- Admin UI integration with update status display
- Background task support for update operations
- Configuration management for update settings

## Test Plan
- [ ] Test version checking against GitHub API
- [ ] Verify update process in development environment
- [ ] Test rollback functionality
- [ ] Validate error handling and logging
- [ ] Review security implications


enferno/commands.py
```diff
@@ -322,3 +428,356 @@ def generate_config() -> None:
...
+        # 6) Restart
+        if restart_service:
+            click.echo("Restarting service...")
+            restart("bayanat")
+            logger.info("Service restarted")
+
+        logger.info("System update completed successfully")
+        return (True, "System updated successfully")
+
+    except Exception as e:
```
tarekio: very generic and catchall. We should instead raise specific exceptions above for each stage, and catch those individually and log them correctly, and then have a catch all at the end.

tarekio (dismissed):
Some issues, and I'm still getting stuck at setup wizard, just need to restart services after install.

enferno/commands.py
level09: Each stage already raises its own specific exception, so the single catch just guarantees the rollback runs every time.  

Splitting it up would mean duplicating that cleanup, or worse, missing a path, because we can’t realistically predict every failure mode. 

Also , the original exception still gets logged in full, so we can investigate the details straight from the logs

enferno/admin/views.py
```diff
@@ -6589,6 +6595,131 @@ def api_bulletin_web_import(validated_data: dict) -> Response:
...
+        return jsonify({"success": True, "message": message})
+    else:
+        current_app.logger.warning(f"System update failed but rolled back: {message}")
+        return jsonify({"success": False, "error": message}), 500
+
+
+@admin.route("/api/system/status", methods=["GET"])
+@auth_required("session")
+@roles_required("Admin")
+def api_system_status() -> Response:
```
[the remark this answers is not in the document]
level09: This endpoint remains from the original flow and acts as a simple “check now” ping that reaches GitHub directly and returns the latest status. 

Redis stays in sync through the Celery job, (and the CLI shortcut I added does the same).

If you prefer that a manual button refresh the shared flag too, I can switch it to call perform_version_check() so Redis updates whenever a manual check is done.

enferno/admin/models/Notification.py
```diff
@@ -172,6 +172,26 @@ def create_for_admins(
 
         return notifications
 
+    @staticmethod
+    def create_for_all_users(
+        title,
+        message,
+        category=NotificationCategories.ANNOUNCEMENT.value,
+        is_urgent=False,
```
tarekio: Set to true

tarekio (commented):
Just a few tiny issues to address.

Separately we should add more integrity checks on update, we can address this in a future PR.

enferno/admin/models/Notification.py
level09: Done

level09:
Closing in favor of an upcoming full rewrite.

PR #207, opened 2025-10-03
fix: cast and strip whitespace for ID number in sheet import

tarekio:
## Jira Issue
1. https://syriajustice.atlassian.net/browse/BYNT-1465

## Description
Cast and strip whitespace for ID number in sheet import

## Checklist
- [ ] Tests added/updated

- [ ] Permissions checked
- [ ] Endpoint tests added

## Additional Notes

### DIFF
--- a/enferno/data_import/utils/sheet_import.py
+++ b/enferno/data_import/utils/sheet_import.py
@@ -660,6 +660,9 @@ class SheetImport:
             if not idn_number or not idn_number.strip():
                 continue
 
+            # cast to string and strip whitespace
+            idn_number = str(idn_number).strip()
+
             # Handle mismatch if type is not valid/missing
             if not idn_type or idn_type not in self.id_number_types:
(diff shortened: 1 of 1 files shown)


enferno/data_import/utils/sheet_import.py
```diff
@@ -660,6 +660,9 @@ def set_idnumbers(self, map_item: Any) -> None:
             if not idn_number or not idn_number.strip():
                 continue
 
+            # cast to string and strip whitespace
```
level09, line 663: I think casting is unreachable here , if `idn_number` is not a string, it will raise an `Attribute error` on line `660`

Maybe better to cast and strip early before the check or maybe a combined check/assignment if you prefer : 

```
  if not idn_number or not (idn_number := str(idn_number).strip()):
      continue
```

tarekio: yep sorry, this was exactly the target of the original issue but missed a push, resolved now.

PR #288, opened 2026-03-16
Add database initialization instructions

djhmateer:
Added instructions for initializing the database.


level09 (commented):
Thanks for catching this missing step! One small suggestion:

```suggestion
Create the database tables, roles, and default data:
```

The current "Create the db tables" is a bit informal and undersells what `create-db` actually does (it also generates roles, workflow statuses, location metadata, and core fields).

djhmateer (commented):
adding in extensions

PR #298, opened 2026-03-23
Improve system configuration organization

apodacaduron:
## Description
Adds a global search box to System Administration to find settings faster.
Selecting a result switches to the correct tab, scrolls to the setting, and highlights it.
Also adds and improves setting grouping titles

## How to Test
1. Open Admin > System Administration.
2. Use the top search box.
3. Select a suggestion and verify it navigates to the right tab, scrolls to the setting, and highlights it.
4. Toggle dark mode and confirm highlight is still visible.
5. Confirm regular tab/settings behavior is unchanged.

## Jira ID (if applicable)
[BYNT-1552](https://syriajustice.atlassian.net/browse/BYNT-1552)

[BYNT-1552]: https://syriajustice.atlassian.net/browse/BYNT-1552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


level09 (commented):
Nice work Daniel, this is a really useful addition — the search + scroll + highlight flow is smooth and well implemented. The Arabic normalization is a great touch too.

A couple of small things before merge:

1. **Missing search entries for a few mail settings**: `MAIL_USE_SSL`, `MAIL_USE_TLS`, `MAIL_USERNAME`, and `MAIL_PASSWORD` aren't in `settingsSearchMap`. Would be good to add them for completeness.

2. **Wrapper divs for IDs**: In some places the `id` could go directly on the existing element (e.g., the `v-switch` or `v-col`) instead of adding a wrapping `<div id="setting-xxx">`. Not a big deal but keeps the DOM cleaner where possible.

Both are quick fixes. Overall this is solid and ready after those tweaks.

apodacaduron:
@level09 Updated, i added pretty much all the missing settings, also did an improvement to show a snack if theres a setting that hidden under another one, for example AWS Region needs that S3 Storage is enabled.

As for "Wrapper divs for ID" i tried but as you know vuetify has many nested divs, so id's would be set on deep elements that prevented the custom background to show consistently, i moved it only on the places that worked fine like v-btn-toggle, h4, v-col, etc

level09:
Tested this locally, works well. The requires handling is the best part, searching for a setting that's hidden behind a disabled parent takes you to the parent and tells you why.

Few questions before we merge:

1. The keywords arrays add around 158 new msgids, mostly single generic words (id, key, on, path, secret, model). Do these need to be translatable at all? The label, section and tab name are already translated and already part of the search, and normalizeForSearch handles the Arabic folding, so Arabic search should still work without them. My worry is that single generic words tend to attract bad fuzzy matches the next time we run translate update.

2. 35 of the 107 entries point at a section heading rather than the setting itself. 11 of the notification ones share setting-notifications-admins-all and 7 of the activity ones share setting-activity-items. So searching "Items deleted" highlights the "Items" heading with six similar toggles underneath. Intentional? Fine if so, it just means the highlight is section level rather than per setting.

3. I noticed a few commits here adding settings that were missing from the map. Worth a dev only console warning in mounted() for any anchorId with no matching element? It passes today, I checked and there are no dead anchors and no unused ids, so it would just catch the next one.

4. Small one, the highlight animation is duplicated at 1579 and 1599, could be a single helper.

Also the results cap is 12 when the box is empty and 24 once you type, was that deliberate?

apodacaduron:
Hey @level09, fixed everything you flagged:

1. Trimmed keywords, un-translated the pure acronyms (s3, aws, csv, oauth, etc). 318 → 112 translated.
2. Gave every setting that shared a heading anchor its own id, so search highlights the right toggle now. Ended up being 8 groups (40 settings), not just the 2.
3. Added a dev console.warn if a setting's anchor is missing from the DOM.
4. Deduped the highlight animation into one helper, matches the form builder's tint now too.
5. 12/24 was an accident, unified to one constant.

Let me know if you catch anything else testing it.

level09 (approved):
Verified the follow-ups: dead-anchor warning covers both anchorId and requires.anchorId (0 dead anchors across all 93), highlight animation deduped into a single helper, results limit unified to one constant, and the shared heading anchors are split out. Keywords trimmed to 112. Remaining shared anchors are parent feature toggles with their children directly beneath, which reads fine.

One note for whoever runs the next translation pass rather than for this PR: the keywords add 70 new msgids, 63 of them single generic words, and pybabel update runs with fuzzy matching on. Worth reviewing those fuzzy entries carefully at that point.

Nice work.

PR #302, opened 2026-03-26
Add Alembic database migrations

level09:
## Summary
- Add Flask-Migrate (Alembic) for database schema migrations
- Baseline revision consolidates all 16 legacy SQL migrations into one idempotent upgrade
- Sync model indexes with migrations (zero schema drift)
- Fix Media.time `Float(precision=2)` to `Float()`
- Fix `normalize_arabic_text()` to include newline collapsing

## Upgrade instructions (for release notes)

**Existing deployments (any version):**
```bash
uv sync
uv run flask db upgrade
```

**New installs:**
```bash
uv run flask create-db --create-exts
uv run flask db upgrade
```

That's it. The baseline migration is fully idempotent. It detects what's already applied and skips it.

## For contributors (going forward)
- Schema changes: `flask db migrate -m "description"`, review the generated file, commit
- Complex migrations: edit the generated file, use `op.execute()` for raw SQL
- Never use `flask create-db` for schema changes after this point

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -71,6 +71,8 @@ export default withMermaid(
             items: [
               { text: "Installation", link: "/deployment/installation" },
               { text: "Configuration", link: "/deployment/configuration" },
+              { text: "Upgrading", link: "/deployment/upgrading" },
+              { text: "Database Migrations", link: "/deployment/migrations" },
               { text: "Docker", link: "/deployment/docker" },
               { text: "Backups", link: "/deployment/backups" },
             ],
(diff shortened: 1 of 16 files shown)


apodacaduron (approved):
- Ran `flask db migrate / upgrade / check` and everything behaved as expected
- Verified `flask db upgrade` is idempotent (safe to run multiple times)
- Tested fresh install flow (`create-db → upgrade → check`) — all good
- Tested upgrade on an existing DB (pre-Alembic) — no duplicate table/index issues
- Confirmed creating invalid migrations fail as expected (no changes applied, fix and try again)

Seems to work well and it's not difficult to use

PR #303, opened 2026-03-27
Add contextual user guide help buttons across mapped pages and dialogs

apodacaduron:
## Description
Adds a reusable contextual help button and wires it to relevant Bayanat docs pages across mapped admin pages/dialogs.
Also adds revision-history help links in entity cards and OCR help in media dashboard/transcription..

## How to Test

1. Open mapped pages/dialogs (Bulletins, Actors, Incidents, Users, Roles, Activity, Import/Export, Media, Dynamic Fields).
2. Confirm help icon appears in toolbar and tooltip says Learn about <topic>.
3. Click help icon and verify it opens the expected docs.bayanat.org/guide/... page in a new tab.

## Jira ID (if applicable)
[BYNT-1560](https://syriajustice.atlassian.net/browse/BYNT-1560)

[BYNT-1560]: https://syriajustice.atlassian.net/browse/BYNT-1560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


level09 (approved):
Looks good. Route-keyed lookup plus one computed in the global mixin is the right amount of machinery for this.

Since it has been open a while I checked the things most likely to have rotted, and they held up: all 18 doc URLs return 200, all 29 route keys match real route names in config.js, and every titleKey resolves to a key that exists in jsapi.jinja2.

One follow-up, not a blocker. Eight of the new strings have no Arabic yet:

Learn about {title}, What is Bayanat?, Activity Monitor, Dynamic Fields, Permissions, Media Management, Video Deduplication, Data Export

The other seven are already translated (Locations, Labels, Sources, Events, Access Control, Media Import, Data Import). Because `Learn about {title}` is one of the missing ones, those seven come out mixed, like "Learn about الأماكن", which reads worse than plain English. #391 adds eight more, so it is one extract and update pass covering both, then Khaleel for the Arabic. Merging this first so it stops aging against main.

Tiny thing whenever you next touch the file: in global-mixin.js currentHelpLabel falls back to `window.translations || {}` and then calls `translations.learnAbout_(title)` on it, which would throw in exactly the case the fallback is there for. It cannot happen in practice since layout.html loads jsapi before the mixin, so either the fallback or the unguarded call can go.

Rebase before merge.

PR #312, opened 2026-04-11
Security hardening for v4 release

level09:
## Summary

- Celery: restrict to JSON-only serialization (remove pickle/msgpack/yaml RCE vector)
- Startup: abort if SECRET_KEY or SECURITY_PASSWORD_SALT not set
- OAuth: fix 2FA bypass via `tf_enter()`, add state parameter for CSRF protection
- Media: add access control to extraction endpoint, fix orphaned file serving with grace period
- Media: add extension validation to inline upload, enforce server-side upload size limit
- Media: replace Referer-based security check with explicit source parameter
- Backup: fix pg_dump string concatenation bug (backups were silently failing) and PATH stripping
- uWSGI: add harakiri timeout (120s) to prevent request-based DoS
- Nginx: add X-Content-Type-Options, Referrer-Policy, Permissions-Policy, X-Frame-Options headers
- SECURITY.md: add v4.x to supported versions

## Test plan

- [x] Full test suite passes (769 passed, 4 skipped)
- [x] Verify media upload flow still works (chunk upload, inline upload)
- [x] Verify extraction endpoint returns data for accessible media, 403 for restricted
- [x] Verify orphaned media files return 404 after grace period
- [x] Verify Celery tasks execute correctly with JSON serialization
- [x] Confirm app refuses to start without SECRET_KEY set

## Summary by CodeRabbit

* **Security**
  * OAuth CSRF state validation added
  * Server-side upload extension and size enforcement (returns 413/415 on violations)
  * Improved media access checks and tightened extraction endpoint auth
  * Added HTTP security response headers
  * Restricted task serialization to JSON
  * Startup now requires secret config; request/worker timeouts added

* **New Features**
  * 2FA integrated into OAuth login flow
  * Explicitly marked support for 4.x release line

* **Bug Fixes**
  * Serve existing on-disk media during grace period when DB record is missing

* **Chores**
  * Added database backup timeout configuration

PR #313, opened 2026-04-12
Modernize deprecated datetime and SQLAlchemy patterns

level09:
## Summary

Mechanical cleanup of long-standing deprecations, dev formatter version pinning with security fix, and two latent bugs uncovered in the process. Zero API, schema, or behavior changes.

### Changes

**Commit 1** — Format untouched files per project black + ruff config (7 files)
- Applied project-canonical formatters to files that had drifted. Pure formatting, no logic changes.

**Commit 2** — Modernize deprecations (47 files)
- `datetime.utcnow()` → `DateHelper.utcnow()` helper. Returns naive UTC (matches existing `db.DateTime` columns exactly — no behavior change, just kills the Python 3.12 deprecation warning).
- `Model.query.get(id)` → `db.session.get(Model, id)` across 162 call sites (SQLAlchemy 2.0 style). 16 files that lacked `from enferno.extensions import db` now import it.
- Pinned `black` and `ruff` in dev dependencies to match pre-commit hook versions (previously `black>=26.3.1` which conflicted with pre-commit; ruff wasn't listed at all).
- Fixed latent circular-import bugs in `Location.py` and `ActorProfile.py` — local deferred imports were critical for class-vs-module name binding during package init, and ruff auto-fix kept removing them. Now marked with `# noqa: F401, F811` + explanatory comments.
- Fixed latent `AtoaInfo` undefined-name bug in `Atoa.query.filter()` path (dead code that would have crashed on first call).

**Commit 3** — Bump black to 26.3.1 (fix GHSA-3936-cmfr-pm3m) (8 files)
- GitHub advisory [GHSA-3936-cmfr-pm3m](https://github.com/advisories/GHSA-3936-cmfr-pm3m) — HIGH severity: "Black: Arbitrary file writes from unsanitized user input in cache file name". Affects all versions `< 26.3.1`.
- Bumped `black==24.10.0` → `black==26.3.1` in both `pyproject.toml` dev deps and `.pre-commit-config.yaml` so they stay in sync.
- Ran black 26.3.1 across the codebase. 5 model files (`Actor`, `ActorProfile`, `Bulletin`, `Incident`, `Media`) picked up minor formatting updates (triple-string collapse inside `db.Computed(...)` and `DDL(...)` calls).
- `uv.lock` updated accordingly.

### Explicitly out of scope (to avoid PR #312 conflicts)

- `enferno/admin/views/media.py`
- `enferno/app.py`
- `enferno/user/views.py`

PR #312 (security hardening for v4 release) actively rewrites these three files and introduces new `.query.get()` calls. A follow-up PR after #312 lands will apply the same cleanup to them plus PR #312's new additions, plus the black 26.3.1 formatting pass on `user/views.py`.

### Test plan

- [x] `uv run pytest` — 769 passed, 4 skipped, 0 failed
- [x] `uv run python -c "from enferno.app import create_app; create_app()"` — imports cleanly
- [x] Live API smoke test against running `flask run` instance at `localhost:5001`:
  - Created bulletin, actor, incident via `POST /admin/api/.../` (201)
  - Updated bulletin with `atob` + `itob` + `btob` relations via `PUT`
  - Verified all rows in DB with microsecond UTC timestamps matching call times
  - Bulletin history + actor history + activity log all written correctly
- [x] Pre-commit hooks pass on all three commits (black 26.3.1 + ruff 0.11.6)


apodacaduron (approved):
System seems to work as usual

PR #315, opened 2026-04-15
Expose OCR provider config in admin dashboard

level09:
## Summary
- Add `OCR_PROVIDER`, `LLM_OCR_URL`, `LLM_OCR_MODEL`, `LLM_OCR_API_KEY` to the system admin dashboard so operators can configure the on-demand OCR backend without editing `.env` and restarting.
- Mask `LLM_OCR_API_KEY` the same way other secrets are handled (`SECRET_FIELDS`, `serialize()`, mask-preserving save), and serialize the four fields through `ConfigManager`.
- Replace the standalone Google Vision key field with a provider-aware block: provider dropdown (`google_vision` / `llm`) with conditional fields for each backend.
- Shrink `.env-sample` OCR section to a short note pointing at the dashboard; the env vars are kept as an optional boot-time override for ops use via the existing `settings.py` fallback.

## Test plan
- [x] Open System Administration → Import and confirm the new "On-demand OCR Provider" block renders with Google Vision selected by default.
- [x] Switch provider to LLM and verify URL/model/API key fields appear; switch back and confirm Google Vision key reappears.
- [x] Save config with LLM fields populated, reload the page, and confirm the API key renders as the masked placeholder (not empty, not cleartext).
- [x] Save again without touching the masked key and confirm the previously stored value survives (no blanking).
- [x] With `LLM_OCR_API_KEY` set in `.env`, confirm it still takes precedence at boot and appears masked in the UI.
- [x] Trigger an on-demand OCR job against both providers and verify extraction still succeeds.


apodacaduron (dismissed):
Approved

Tested the config flow with both .env overrides and system settings, and it behaved as expected. The .env values still take precedence at boot, which matches the intended fallback behavior from the test plan.

Only small nit: if OCR_PROVIDER is set in .env to an invalid value, like google_visionnn, that invalid value still shows up in the provider select. It could be safer to validate the env value and fall back to a supported default instead of reflecting an unknown option in the UI.

level09:
Thanks for the review @apodacaduron! Good catch on the invalid `OCR_PROVIDER` showing up in the dropdown.

Addressed in bfd71cd99: `settings.py` now coerces unknown values to `google_vision` at boot and logs a warning, so a typo like `google_visionnn` in `.env` no longer reaches the dashboard select. Validated against the canonical `PROVIDERS` set from `enferno/utils/ocr/__init__.py` to keep a single source of truth.

PR #321, opened 2026-04-21
Pentest fixes and hardening

level09:
Fixes for the 7ASec BAY-01 engagement, on top of baseline `8615ff11b`. One commit per finding, tagged `fix(BAY-01-NNN):` so the retest maps fixes by `git log --grep BAY-01 pentest-fixes`.

42 of 43 fixed. Open: `BAY-01-017` (release signing), a product decision, not a code gap.

## Finding to commit

```
001 80f56b9bc   015 4f8b40e68   028 0cddfca22   039 aa7ecb076
002 d0f4581da   016 18059510d   029 bc467b9df   040 9bfca8af5
003 3ce769fe8   018 ce2afebcb   030 fa3069dcd   041 726d34582
004 4b66981d5   019 ee18f11ee   031 dc8e62fec   042 b186e1b96
005 597e7e6c1   020 f55f7a01c   032 adf941e8d   043 c0adefa7c
006 507bab008   021 f541199f7   033 5bbacd183
007 ecc4aa76f   022 85e7f540a   034 279b3c8f8
008 9800fbed6   023 0cba2461f   035 f1990806b
009 0614f8aad   024 09f3f60b5   036 5abe376ea
010 f34e6790c   025 4708fe422   037 9ee85af9f
011 dec0516ea   026 76f431fff   038 6f562d461
012 8b0c2d5c3   027 a5280d017
013 e4f26caac   014 ddabe8eaa
```

017 is unaddressed pending the signing decision. minisign keypair exists, public key pinned; manual `sudo bayanat update` stays as the fallback.

## Partials worth flagging

- `020`: filename enumeration closed (opaque names). Per-item auth on `/api/serve/inline/<filename>` is a separate parent-binding change, deferred.
- `028` issue 6: passwords are out of argv (in-container config file + `REDISCLI_AUTH`). Delivery stays via env vars because Compose can't mount env-sourced secrets into `read_only` containers.
- `031`: app DB role stripped of superuser/createdb/createrole/replication and CONNECT revoked from PUBLIC, but keeps table ownership because dynamic-fields runs `ALTER TABLE` at runtime.

## Retest

Backend regression tests: `uv run pytest tests/test_pentest_fixes.py`. No frontend harness in this repo; 034 verified by review. Each commit message has root cause and where to verify. Fresh install recommended (005 changed the install flow: the installer now generates and prints the admin password once).


apodacaduron (commented):
**BAY-01-001:**
- If the user does not belong to the same access group as a restricted bulletin, the history endpoint correctly blocks the request: `GET /admin/api/bulletinhistory/3` returns `{"message": "Restricted Access"}` ✅
- Direct access to the restricted bulletin also correctly blocks: `GET /admin/api/bulletin/3` returns restricted access and triggers the admin notification ✅
- Related follow-up: the relations endpoint still returns `200` for the same restricted parent bulletin. Example: `GET /admin/api/bulletin/relations/3?class=actor` returns masked entity stubs plus relationship metadata like `related_as`, `probability`, `comment`, and `user_id`. This leaks relationship existence/metadata for a restricted parent record and should probably return `403`. ⚠️
- Related follow-up: the graph endpoint also returns `200` for the same restricted bulletin. Both `GET /admin/api/graph/json?id=3&type=bulletin&expanded=false` and `GET /admin/api/graph/json?id=3&type=bulletin&expanded=true` return masked nodes, but still expose graph topology and metadata, including restricted entity IDs, relation existence, relation type (`Perpetrator`), connected accessible records, and connected locations (`Mexico`, `USA`). This should probably return `403` when the root entity is restricted from the current user. ⚠️
- Related follow-up: the flowmap endpoint masks the restricted actor but still leaks event/location timeline data. Example: `POST /admin/api/flowmap/actors-for-locations` with `{"location_ids":[1],"q":[{}]}` returns actor `1` as restricted, but includes `current_event` and `next_event` details such as event type (`Detained/Abducted`, `Death`), locations (`Mexico`, `USA`), location IDs, and event dates. Restricted actors should probably be excluded from the flowmap result or returned without event details. ⚠️

apodacaduron:
## BAY-01-002
- Unauthorized DA: `PUT /admin/api/extraction/1` → `403` ✅
- Authorized user: `200` with compact response only — no `text`, `original_text`, or `history` leaked ✅
- `cant_read` action also blocks unauthorized DA → `403` ✅

apodacaduron:
## BAY-01-003

- Requested export as non-admin DA with a mix of accessible and restricted bulletin IDs ✅
- Approved as admin, downloaded CSV, PDF, and JSON ✅
- Restricted bulletin absent as a direct export item across all 3 formats ✅
- Related follow-up: restricted records still leak via `bulletin_relations`, `actor_relations`, and `incident_relations` on accessible exported items — `to_dict()` serializes all relations unconditionally with no access check. Not covered by this fix ⚠️
- Related follow-up: `/export/dashboard/<id>` details drawer for incident exports sends `q` as an array instead of an object — `POST /admin/api/incidents/` returns `400` validation error. Bulletins and actors use an array correctly, incidents expect an object (`IncidentQueryModel`). ⚠️

apodacaduron:
## BAY-01-004

- `POST /import/api/csv/analyze` with `../../../../etc/passwd` → `400` ✅
- `POST /import/api/xls/sheets` with `../../../../etc/passwd` → `400` ✅
- `POST /import/api/xls/analyze` with `../../../../etc/passwd` → `400` ✅
- `POST /import/api/csv/analyze` with `../../../../app/.env` → `400` ✅ (confirmed leaked `SECRET_KEY` and secrets on `main`)
- Legitimate filename inside `IMPORT_DIR` still parses normally ✅

apodacaduron:
## BAY-01-005
- `POST /api/create-admin` → `404` ✅
- `GET /api/check-admin` → `404` ✅
- `GET /setup_wizard` unauthenticated → redirects to `/login` (requires `@auth_required`) ✅

apodacaduron:
## BAY-01-007
- `POST /login` with bad credentials in a loop → `429` after 10 attempts ✅
- `X-RateLimit-Limit: 10`, `X-RateLimit-Remaining: 0`, `Retry-After: 880` headers present ✅

apodacaduron:
## BAY-01-008:

- The original reported import path for Actor Profile Description appears to be fixed: importing `` through the sheet import stores a sanitized value, and the `onerror` handler is removed before the value reaches the actor description `v-html` sink ✅
- Follow-up found while testing the same import surface: the Missing Person import path still persists raw HTML when MP fields are mapped/populated and the actor profile is created with `mode = 3`. The payload is then rendered by `MissingPersonField.js` via `v-html`. In my test, CSP blocked execution of the `onerror` handler, but the raw attacker-controlled HTML was still stored and rendered, so this should not be considered fully remediated ⚠️
- Suggested fix: apply the same `sanitize_string()` policy to imported MP fields before persistence, especially MP detail fields such as `seen_in_detention_details`, `injured_details`, `known_dead_details`, and `skin_markings_details`, plus direct MP text fields rendered by `MissingPersonField.js`. Alternatively, render these values as plain text instead of `v-html`.

apodacaduron:
## BAY-01-009:

- The main media update endpoint looks fixed now. When I tested `PUT /admin/api/media/<id>` as another DA user who can see the bulletin but is not assigned to it, the request correctly returned `403` ✅

- However, I found the same assignment issue still exists on related media actions. As the unassigned DA user, these still returned `200`:
  - `PUT /admin/api/media/<id>/orientation`
  - `POST /admin/api/ocr/process/<media_id>`
  - `POST /admin/api/ocr/bulk`
  - `PUT /admin/api/extraction/<extraction_id>`

- So the direct media update is fixed, but the related media/OCR/extraction actions still seem to be missing the same assigned-user check. A DA who is not assigned to the bulletin can still rotate media, trigger OCR, and update extracted text on another analyst’s assigned item ⚠️

- I think these endpoints should use the same edit boundary as `PUT /admin/api/media/<id>`, and we should add regression tests for the related media actions too.

apodacaduron:
## BAY-01-010
- The request completed, and the restricted/private relation targets were skipped and were not added to the Bulletin ✅

apodacaduron:
## BAY-01-011:
- Tested the Moderator bulk update case Bulletin access roles were not cleared after processing ✅

apodacaduron:
## BAY-01-012:
- Direct media/file access is now blocked with `403` / `Unauthorized` ✅
- The transcribe/media API flow is also blocked with `403` ✅
 
NOTE: the Bulletin drawer still renders the Media card for this user, including media ID, filename, etag/hash, and media action buttons. The file itself is blocked, but media metadata is still exposed through the parent Bulletin view/UI

apodacaduron:
## BAY-01-035:

- Tested the ragged CSV reproducer (`a,b,c` followed by `1,2,3,4,5`) against `SheetImport.parse_csv()`.
- The parser now returns a normal JSON-serializable structure instead of tuple keys ✅
- The extra fields are dropped and the result serializes cleanly, so this looks fixed.

apodacaduron:
## BAY-01-036
- Uploading a .xls file returns HTTP 500 Internal Server Error ❌
- The fix pins engine="openpyxl" but openpyxl can't read .xls files, so the error still happens, just differently
- .xls is still in the allowlist so users can still upload it

apodacaduron:
## BAY-01-037
- sheet=null and sheet=[0] on /import/api/xls/analyze correctly return 417 ✅
- sheet=true on /import/api/xls/analyze returns HTTP 500 ❌
- sheet=true on /import/api/process-sheet returns HTTP 500 ❌ — this endpoint has no type check

apodacaduron:
## BAY-01-038:

- XLSX import works correctly with text headers (`a`, `b`, `c`) mapped to actor name fields ✅
- With numeric headers (`1`, `2`, `3`), the mapping screen accepts them, but the imported actor name fields are not populated and the actor is created as `UNKNOWN UNKNOWN` ⚠️

apodacaduron:
## BAY-01-039:
- Covered while testing BAY-01-008: the main sheet-import description and mismatch paths now sanitize the XSS payloads before they reach the actor description ✅
- Related follow-up remains in the Missing Person import fields, which can still render raw imported HTML through `MissingPersonField.js` with CSP blocking execution ⚠️

apodacaduron:
## BAY-01-040:
- Tested malformed export request bodies like `null`, `[]`, `"hello"`, `{}`, `{"config": null}`, and `{"config": "foo"}`.
- They no longer crash the server with HTTP 500; the API now returns a controlled `{"message": "Error creating export request"}` response ✅

apodacaduron:
## BAY-01-042:

- Tested the malformed import API bodies from the report. Those now return controlled responses instead of HTTP 500 ✅
- Follow-up: `/import/api/process-sheet` still returns HTTP 500 for malformed `files` input, for example `{}`, `{"files": null}`, or `{"files": "x"}`. This looks like the same class of issue and should validate that `files` is a non-empty list before processing ⚠️

apodacaduron:
BAY-01-015 looks fixed.

I tested by changing the export ID in the URL/API to another user’s export. The server returned 403 Forbidden, so the export metadata is no longer exposed.

Minor note: 403 vs 404 can still hint that an export ID exists, but that is low impact and not the main issue.

Unrelated UX issues noticed while testing:
- If an Admin changes their own role from Admin to Mod while on /admin/users/, they land on a raw Forbidden page with no way out. They have to manually change the URL to a valid path.
- As a Mod, /admin/api/updates/available returns 403 and shows a “You do not have permission to view this resource” snackbar.

apodacaduron:
BAY-01-018 looks fixed for the main reported issue.

I tested a bulk Bulletin update as a Mod with:
- Bulletin 12: accessible
- Bulletin 11: restricted

Only Bulletin 12 was updated. Activity Monitor only showed Bulletin 12, and only Bulletin 12 got a new revision history entry.

Separate issue found while testing Incident bulk updates:
When using the assignRelated option, related Bulletins can get duplicate revision history entries.

Test:
- Set chunk size to 1 for testing
- Bulk updated 2 incidents with assignRelated=true
- Incident 1 was related to Bulletin 12
- Incident 2 was related to Bulletins 12, 11, and 10

Result:
Bulletin 12 received 3 new revision history entries from one bulk operation.

So the original restricted-ID logging issue seems fixed, but assignRelated can still create duplicate history entries for related Bulletins.

apodacaduron:
BAY-01-022 looks fixed for the normal Bulletin update API, but I found a remaining bulk-update edge case.

Test:
- Bulletin 10 was assigned to a DA user.
- Its status was set to Peer Review Assigned.
- As the assigned DA user, a direct normal update returned 403: {"message": "Item is locked for peer review"}.
- However, using a Mod account, the bulk update endpoint still allowed updating Bulletin 10 while it was in Peer Review Assigned.

So the normal edit path is protected, but bulk update can still mutate peer-review-locked Bulletins.

apodacaduron (dismissed):

apodacaduron (approved):

PR #325, opened 2026-05-02
Feat/custom s3 endpoint

adelsolt:
# Adding a custom S3 endpoint support for non-AWS providers

While deploying Bayanat for a human rights project, we needed to use OVH Object Storage instead of AWS. boto3 works fine with any other providor but without an endpoint, it silently goes to Amazon and rejects everything.

The codebase had no way to configure this. Three small changes fix it:

## Changes

**1) `enferno/utils/config_utils.py`**:  `S3_ENDPOINT` needs to be added in 3 places inside this file:
 
- **Default values block**:  so the key exists with an empty default:
```python
"AWS_REGION": "",
"S3_ENDPOINT": "",
```
 
- **UI labels block**:  so it shows up correctly in the setup wizard:
```python
"AWS_REGION": "AWS Region",
"S3_ENDPOINT": "S3 Endpoint URL",
```
 
- **Config reader block**:  so the current value gets read back:
```python
"AWS_REGION": cfg.AWS_REGION,
"S3_ENDPOINT": cfg.S3_ENDPOINT,
```

**2) `enferno/settings.py`** — one line added to load `S3_ENDPOINT` into Flask's `app.config`:
```python
S3_ENDPOINT = manager.get_config("S3_ENDPOINT")
```
Without this, `current_app.config.get("S3_ENDPOINT")` returns `None` everywhere even if `config.json` has it.

**3) `enferno/admin/views/media.py`**: added `endpoint_url` to all 4 boto3 calls (`_media_url`, `api_medias_chunk`, `api_medias_upload`, `serve_media`):
```python
endpoint_url=current_app.config.get("S3_ENDPOINT")
```

## Backward compatibility

If `S3_ENDPOINT` is not set, `config.get()` returns `None` and boto3 falls back to AWS as before, nothing breaks for existing deployments.

### DIFF
--- a/enferno/admin/views/media.py
+++ b/enferno/admin/views/media.py
@@ -103,6 +103,7 @@ def _media_url(media_file):
         aws_access_key_id=current_app.config["AWS_ACCESS_KEY_ID"],
         aws_secret_access_key=current_app.config["AWS_SECRET_ACCESS_KEY"],
         region_name=current_app.config["AWS_REGION"],
+        endpoint_url=current_app.config.get("S3_ENDPOINT"),
     )
     return s3.generate_presigned_url(
         "get_object",
@@ -237,6 +238,7 @@ def api_medias_chunk() -> Response:
                 aws_access_key_id=current_app.config["AWS_ACCESS_KEY_ID"],
                 aws_secret_access_key=current_app.config["AWS_SECRET_ACCESS_KEY"],
                 region_name=current_app.config["AWS_REGION"],
+                endpoint_url=current_app.config.get("S3_ENDPOINT"),
             )
             s3.Bucket(current_app.config["S3_BUCKET"]).upload_file(filepath, filename)
             # Clean up file if s3 mode is selected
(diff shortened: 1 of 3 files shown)


level09:
Hi @adelsolt, thanks for the PR.

This works with zero code change. Bayanat is on `boto3` 1.42 / `botocore` 1.38, which reads `AWS_ENDPOINT_URL_S3` from the environment automatically (since botocore 1.34). One line in `.env`:

```
AWS_ENDPOINT_URL_S3=https://s3.gra.io.cloud.ovh.net
```

routes every boto3 callsite through the custom endpoint. Verified against our exact version.

The env-var path also covers four other boto3 callsites the PR doesn't touch (`tasks/exports.py`, `tasks/extraction.py`, `utils/backup_utils.py`, `data_import/utils/media_import.py`), so going with code changes here would leave exports, OCR, backups, and imports still pointing at AWS.

Closing this and adding a documented line to `.env-sample` instead. Thanks for the heads up.

PR #332, opened 2026-05-07
docs: add db init steps to README quick start

level09:
The Quick Start section skips schema creation, so `flask install` exits with an error on a fresh clone (no Role table to query).

Add the two documented init steps between `gen-env.sh` and `flask install`:

- `flask create-db --create-exts` creates the schema and PostGIS / pg_trgm extensions
- `flask db stamp head` marks Alembic at the current revision so a later `flask db upgrade` is a no-op


apodacaduron (commented):
Should we also add `uv run flask db stamp head` to `docs/deployment/installation.md` so the full install guide matches this Quick Start flow?

apodacaduron (dismissed):

level09:
Good catch, thanks. Already covered in e32a6dfb1, installation.md now has `uv run flask db stamp head` right after `create-db`. Your review just landed on the older commit.

apodacaduron (approved):

PR #335, opened 2026-05-11
Make session idle timeout configurable via SESSION_LIFETIME env

level09:
## Summary
- `PERMANENT_SESSION_LIFETIME` is currently hardcoded to 3600s. This makes it configurable via the `SESSION_LIFETIME` environment variable, defaulting to 3600 (no behavior change for existing installs).
- Lets operators tune the idle session timeout (Flask refreshes the cookie/Redis TTL on every request, so this effectively controls idle expiry) without code changes.

## Test plan
- [ ] Unset `SESSION_LIFETIME` → sessions expire after 3600s of inactivity (unchanged default).
- [x] Set `SESSION_LIFETIME=300` in `.env` → sessions expire after 5 min of inactivity.
- [x] Service starts cleanly with the env var present and absent.


apodacaduron (approved):
Simple change works as intended

PR #337, opened 2026-05-12
fix(auth): clear force-reset flag when password is written outside web flow

level09:
## Summary

Admin-triggered force-password-reset uses a Redis flag (`security:user:<id>`) checked by a global `before_app_request` guard that redirects every request to `/change` until the user updates their password. The flag is cleared by Flask-Security's `password_changed` signal, but that signal only fires through the web `/change` form. CLI password resets (`flask reset`, `flask reset-all-passwords`) and the admin user-edit endpoint write a fresh hash directly and skip the signal, so the flag persists. The user then sees a successful login but is stuck in a redirect loop because the flag never clears.

## Fix

- Add `User.set_password(password)` which hashes the password and clears the force-reset Redis flag in one step.
- Route the CLI `reset` and `reset_all_passwords` commands and `User.from_json` through it.
- `reset_all_passwords` still calls `set_security_reset_key()` afterward, since its purpose is to issue a temporary password and force the user to change it.
- New-user paths (`install`, `create`) are unchanged. They build a `User` before save, so no `id` and no flag can exist.

The web `/change` flow is untouched and continues to clear the flag via the existing `password_changed` signal handler.

## Test plan
- [x] Set the force-reset flag for an existing user, then run `flask reset -u <user> -p <new>` and confirm the user can access non-`/change` pages on next login.
- [x] Confirm `flask reset-all-passwords` still leaves the force-reset flag set so all users must change their passwords on next login.
- [x] Confirm an admin editing a user's password via the admin UI no longer leaves the flag dangling.
- [x] Web `/change` flow still clears the flag (regression check).


apodacaduron (dismissed):
Tested locally — Redis flag gets cleared correctly from CLI reset, admin UI, and reset-all-passwords behaves as expected too, no regressions

apodacaduron (approved):
Checked again and fix still working, flag removed when changing password from cli

PR #343, opened 2026-05-23
Exclude notification poll from sliding the idle session timeout

level09:
## Problem

The session idle timeout (`PERMANENT_SESSION_LIFETIME`) never fires. The notification poller hits `GET /admin/api/notifications` every 60s, and with Flask-Session's default `SESSION_REFRESH_EACH_REQUEST=True` each request rewrites the Redis session TTL back to full lifetime. A user who leaves a tab open is never logged out.

## Fix

Mark the automatic poll with an `X-Silent-Poll` header and override `should_set_storage` so requests carrying it don't refresh the session expiry (unless they actually modified the session). All other requests, including user actions in the notification drawer, still slide the session.

- `enferno/app.py` — per-request opt-out in `should_set_storage`.
- `enferno/static/js/mixins/notification-mixin.js` — only the interval-driven `refetchNotifications` sends the header; user-initiated loads do not.

## Tests

`tests/test_session_poll_refresh.py`: silent poll leaves the TTL untouched; a normal request still refreshes it.


apodacaduron (approved):
Works as intended, notifications endpoint no longer refreshes session

PR #305, opened 2026-04-05
Add public archive export command

level09:
## Summary

- New `flask export public` CLI command that exports labeled bulletins as denormalized JSON for the Bayanat Document Archive (BDA) portal
- Serializes only public-facing fields, strips all workflow/internal metadata
- Copies media files from local filesystem or S3 (auto-detected from config)
- Eager-loads all relationships to avoid N+1 queries

## Usage

```bash
flask export public --label "public-archive" --output ./export/
flask export public --label "public-archive" --output ./export/ --no-copy-media
```

Produces `documents.json` + `media/` directory consumed by the [bayanat-archive](https://github.com/sjacorg/bayanat-archive) portal.

## Fields exported

Bulletin: id, title, title_ar, description, source_link, publish_date, documentation_date

Relations (denormalized): labels, verified_labels, sources, locations, geo_locations, events, media (with OCR extraction text), related_bulletins, related_actors, related_incidents

## Fields excluded

status, assigned_to, user_id, peer reviewers, comments, review, review_action, tags, originid, meta, tsv, sjac_title, sjac_title_ar, roles, reliability_score, dynamic_fields

## Test plan

- [x] Tested with 323 bulletins on local filesystem (25 media files copied, 0 missing)
- [x] Tested with 10 bulletins on S3 storage (10 media files downloaded, 0 missing)
- [x] Verified zero internal fields leaked in output JSON
- [x] Verified all existing CLI commands still registered and functional
- [x] Error handling: nonexistent label exits with clear message

PR #345, opened 2026-06-01
Add public archive export with public_description field

level09:
Adds the `flask export public` command that exports labeled bulletins as denormalized JSON plus media for the public archive.

Also adds a curated `public_description` field on bulletins: editable in the bulletin form (core dynamic field) and exported as the public-facing context for the archive, kept separate from the internal `description`. Includes the Alembic migration that adds the column and registers the field.

Supersedes #305 (same export command, rebased onto current main).


apodacaduron:
`public_description` is not shown automatically in the UI because core fields are rendered through explicit `v-if` conditions, so they can use custom components/layouts instead of the generic `field-renderer`.

Should I add `public_description` to the bulletin UI using the TinyMCE editor, similar to the existing `description` field?

apodacaduron:
Quick security question: right now the export seems to only use the label to decide what goes public. Should we also make sure restricted bulletins, or restricted related records, are excluded?

I’m asking because if something restricted gets the label by mistake, this could leak information. Maybe we should handle it similar to the penetration-test fixes?

level09:
Yes, let's use TinyMCE like `description`. Go ahead and add it to the bulletin form right after `description`. Thanks for checking!

level09:
Good catch. Let's keep it out of this PR though, since the export is admin-only CLI there's no privilege escalation, and the label is our curation gate for what gets exported. The actual "safe to publish" review will happen on the archive side: documents import as drafts and we publish them per-document after seeing the rendered output. The restriction-leak hardening is already tracked with the pentest fixes.

apodacaduron:
> Yes, let's use TinyMCE like description. Go ahead and add it to the bulletin form right after description. Thanks for checking!

@level09 On it 🫡

apodacaduron (approved):
I updated the UI so `public_description` renders next to the existing `description` field and uses the same TinyMCE editor. I also added visible labels for both editors, since having two TinyMCE fields stacked together was ambiguous.

For fresh installs, I re-ordered the core fields so `public_description` comes right after `description`.

For existing DBs, the migration now gives `public_description` the same `sort_order` as `description` instead of shifting the rest of the fields. That keeps it close to `description` without changing customized form layouts. Admins can still reorder it in the form builder if they want a precise position.

PR #348, opened 2026-06-09
Remove retireignore file and update TinyMCE

apodacaduron:
## Description
We added this file because after we updated tinymce #347 pipeline was still failing, this was due to some outdated rules in retirejs package, they sent a fix so this file is not needed anymore

## How to Test
RetireJS pipeline should pass

## Jira ID (if applicable)
[e.g., BYNT-123]


level09 (dismissed):

apodacaduron:
I updated Tinymce to 8.7.0 but seems like the updated dompurify version will be release on 8.7.1 so this will remain blocked for now https://github.com/tinymce/tinymce/milestone/103?closed=1

apodacaduron:
@level09 will close this as almost every new release of dompurify fixes some xss vulnerability and a few days later a new one is found, will close this until it stabilizes a bit

PR #349, opened 2026-06-09
Document & image redaction

level09:
## Summary
In-app visual redaction for document and image media. An analyst draws boxes over a PDF or scan, Bayanat burns them in permanently and saves a clean redacted copy as a **new** Media row. The original is never modified.

## How it works
- **Engine** (`enferno/utils/redaction_utils.py`): PDFs via PyMuPDF `add_redact_annot` + `apply_redactions` + `scrub` (destructive: content removed, not overlaid); images via Pillow flatten to JPEG (EXIF dropped).
- **Endpoint** `POST /api/media/<id>/redact` (Admin/DA): reads source bytes, burns, writes a new file, creates the Media row plus a `media_redaction` audit row (source, regions, user), logs an Activity. Existing redactions can also be deleted (#354).
- **Available on bulletins and actors**: redact icon on the media card next to the OCR icon.
- **Many copies of one doc**: redacted copies are grouped by their source media id and collapsed behind a "Show N redactions" toggle, so a document redacted once per detainee does not clutter the record. Optional per-copy label keeps them findable; copies can be edited in place.
- **Migrations**: `e4f2a9c8d7b1` (additive `media_redaction` table) and `68396035f041` (`original_media_id` link). Both additive with clean downgrades, no data risk.
- **Docs**: Redaction guide included.

## Status
Ready for review and merge. UX has been finalized with the team and the flow has been validated on staging (sta.bayanat.org).

## Review focus
- Irreversibility: PDF text/content actually gone under boxes; scanned image-only pages blanked per region (not the whole page); image EXIF dropped.
- Original Media row/file immutability (redaction always produces a new row).
- Access control on the endpoint (`@roles_accepted("Admin","DA")` + `can_access`).


apodacaduron (dismissed):
This is working as expected, permissions match the OCR feature so it should behave the same as that's in terms of availability for certain roles, also added the redaction tool to actor drawer / editor, one small note is that the delete feature its implemented too but the backend work for that is missing, the endpoint needs to be added to be able to delete a redaction

apodacaduron (dismissed):
Delete works now! Good to go

apodacaduron (approved):

PR #355, opened 2026-06-27
Add independent incident scope for event types

level09:
## Summary
Event types now have an independent **Incidents** scope, separate from Bulletins. Previously the Incidents screen reused the `for_bulletin` flag, so an event type assignable to Bulletins was automatically assignable to Incidents, with no way to scope a type to Incidents alone. This adds a third toggle alongside the existing Actors and Bulletins ones, mirroring the established `for_incident` pattern already used by Labels.

## How it works
- **Model/API/validation**: new `Eventtype.for_incident` column, serialized in `to_dict`, accepted in `from_json` and the validation model; the eventtypes API now accepts `typ=for_incident`.
- **Admin editor**: a "For Incidents" checkbox (three equal columns: Actors / Bulletins / Incidents) and a matching table column.
- **Incident consumers** point at the new flag instead of `for_bulletin`: the event picker (`incidents.html`) and the search/filter sidebar (`IncidentSearchBox.js`), matching how the Actor and Bulletin search boxes use their own flags.
- **Seed data** (`enferno/data/eventtypes.csv`): new `for_incident` column, set equal to `for_bulletin` per row so fresh installs match the migration backfill.

## Migration
`c7d2e9f4a1b8` adds the column and backfills `for_incident = for_bulletin` for existing rows, so the Incidents event-type list is unchanged on day one. Admins can then curate it independently. No data is removed; existing incident events keep their types regardless of the flag.

## Tests
Eventtype CRUD test exercises the new field; factory and `create_eventtype_for` helper extended to support incident-scoped types. All eventtype lookup tests pass.


apodacaduron (approved):
Working well, matches same values of bulletin after migration, setting new event types show correctly on incidents and it's searchable as needed

PR #358, opened 2026-07-07
Redesign Add/Edit User dialog and add role descriptions

apodacaduron:
## Description
Redesigns the Add/Edit User dialog into clearly labeled sections (User Details, Account Status, User Permissions), moves the field-enable pencil icons inline into their inputs, restyles the Group multi-select dropdown to show a description under each role with a checkmark/highlight for selected ones, and adds real descriptive text for the default Admin/DA/Mod roles (replacing the generic "System Role" placeholder) via a data migration. No functional/permission behavior changed — same fields, validations, and API calls as before.

## How to Test
1. Go to **Admin > Users**.
2. Click **New User** — confirm the dialog shows sectioned layout (User Details / Account Status / User Permissions), the "Add User" button, and the Group dropdown shows role name + description with a checkmark on selected roles.
3. Edit an existing user — confirm the pencil icons next to Email/Username/Password are clickable even though those fields start disabled, and the footer button reads "Save" with a checkmark icon.
4. Toggle a few permission switches and roles, save, and confirm the change persists.
5. Run `flask db upgrade` and confirm Admin/DA/Mod roles get updated descriptions (only rows still set to "System Role" are touched).


level09:
Pushed a small fix to the migration on this branch (0bdd5b8af), since it is easy to miss on the FE side.

`479c8b98b036` was parented on `68396035f041`, which stopped being the tip of the chain once the redaction and incident branches got merged into `f0a3d6c1e8b2` on main. After the last main merge this branch had two Alembic heads, so `flask db upgrade` would have failed on deploy with "Multiple head revisions are present". Re-parented it onto `f0a3d6c1e8b2`, back to a single head. Black also reformatted the file on commit.

The migration itself is unchanged, it is data-only so ordering does not matter. Mind giving it a quick look when you get a chance?

One thing to watch: whichever migration PR merges first wins the head. If #370 lands before this one, this will need re-parenting onto `b6e3d1a8f254`.

PR #359, opened 2026-07-09
Mirror actor relations and convert type on create/update

apodacaduron:
## Summary
Actor relations weren't mirroring correctly. If you related actor A to actor B as "Parent," both actors would show "Parent" instead of B correctly showing "Child." Depending on which actor you created the relation from, the relation could also get saved backwards in the database, not just displayed wrong.

This is now fixed for all 5 directional relation types (Parent/Child, Superior/Subordinate officer, Unit/Subunit, Alleged Perpetrator/Injured Party, Member/Group), on both creating a new relation and editing an existing one's type — including editing from the bulletin or incident form, and relating during actor creation before the new actor has an id. Symmetric types (Sibling, Spouse, Same Person, Duplicate, Other) and cross-type relations (Actor↔Bulletin, Actor↔Incident) are unaffected — no mirroring is needed for those since they never had this problem.

No backend or database changes — this is a frontend-only fix. Storage stays exactly as it was; the correct label is computed based on which actor you're viewing, and the correct type is saved based on which actor you're relating from.

## Test plan
- [x] Create two actors, note which has the lower id and which has the higher id.
- [x] Relate them as Parent/Child, starting from the lower-id actor.
  - [x] Lower-id actor shows "Parent".
  - [x] Higher-id actor shows "Child".
- [x] Now do the same starting from the higher-id actor instead — pick "Parent" from there.
  - [x] Confirm it still comes out correct on both sides.
- [x] Repeat for at least one more directional pair (e.g. Superior officer/Subordinate officer).
- [x] Edit an existing relation's type (e.g. change Parent to Child) — confirm both actors update correctly.
- [x] Create a brand new actor, relate an existing actor as Parent during creation (before saving), then check both profiles after save.
- [x] From the bulletin edit form, change a related actor's relation type (e.g. Witness to Perpetrator) — confirm it saves exactly what you picked, not a different type.
- [x] Sanity-check a symmetric type (e.g. Sibling) still shows the same on both sides.
- [x] Sanity-check Actor↔Bulletin and Actor↔Incident relations are unaffected when related normally (no type change).

## Jira ID (if applicable)
[BYNT-1721](https://syriajustice.atlassian.net/browse/BYNT-1721)

[BYNT-1721]: https://syriajustice.atlassian.net/browse/BYNT-1721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/enferno/admin/templates/admin/actors.html
+++ b/enferno/admin/templates/admin/actors.html
@@ -1637,6 +1637,12 @@
                                 this.editedItem.comments = '';
                                 this.resetSwitchers();
                                 this.handleStatus(item);
+                                this.applySameTypePerspective({
+                                    relationList: this.editedItem.actor_relations,
+                                    type: 'actor',
+                                    editedItemId: this.editedItem.id,
+                                    sameTypeInfo: this.atoaInfo,
+                                });
                             });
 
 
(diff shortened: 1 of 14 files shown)


level09 (approved):
Excellent. 

Let's get https://github.com/sjacorg/bayanat/pull/360 approved and sync this to fix the pipeline.

PR #362, opened 2026-07-11
Export all media files per item instead of only the first

level09:
Fixes BYNT-1727.

`generate_export_media` copied only `item.medias[0]`, so bulletins (and actors) with multiple media attachments exported incomplete data. Now all media files linked to each exported item are included, for both local filesystem and S3 storage. The per-file throttle sleep moved inside the media loop so S3 download pacing stays consistent.

Media filenames are UUID-based so there are no collision concerns in the export folder.


apodacaduron (approved):
Tested using local filesystem it does bring back all related files now

PR #364, opened 2026-07-14
Remove OCR text map overlay and stop storing raw OCR payloads

level09:
## What

The OCR text map overlay turned out to serve a very narrow case: it renders only on single image files (not PDFs) and only with the Google Vision provider (the LLM provider returns no geometry). Meanwhile its fuel, the full provider responses stored in `extraction.raw`, accounts for 37 GB on large deployments, most of it per-character coordinates nothing ever read. Decision: remove the tool and stop storing raw provider output entirely. If a future feature needs text geometry, it can re-run OCR on the specific document on demand.

## Changes

- Remove `OcrTextLayer.js` and its registrations (bulletins, actors, media-dashboard templates); remove the Text Map toggle and overlay branch from `MediaTranscriptionDialog.js`.
- Stop storing `raw` on successful extractions; providers (Google Vision, LLM, DOCX) no longer return it. Failed extractions keep their small error payload for debugging.
- Stop attaching `raw` to media API responses.
- New `flask ocr purge-raw` command (opt-in, NOT a migration): clears stored payloads from existing rows in batched transactions, resumable, with `--dry-run`. Failed-extraction error blobs are preserved.
- The `raw` column stays; no schema change, no migration. Installs that never run the purge are unaffected.

## Notes for operators

- Optionally archive the extraction table first (`pg_dump -Fc --table=extraction`) if you want to retain old provider output.
- `flask ocr purge-raw --dry-run` reports how many rows hold payloads.
- Disk space returns to the OS after `VACUUM FULL extraction` (locks table) or pg_repack.

Supersedes the earlier slim-only approach (first commit); the second commit removes the feature and the write path entirely.


level09:
Superseded by a clean single-commit PR: the branch here accidentally included unrelated history (100 commits), making review impossible. Same change, rebuilt on main, in the replacement PR.

PR #365, opened 2026-07-15
Search translated titles in lookup typeahead endpoints

level09:
## Summary
Typeahead search on several lookup endpoints only matched the primary `title`, so typing a translated (Arabic) term returned no suggestions even when `title_ar` was populated. This brings them in line with the reference-data endpoints (countries, ethnographies, dialects, ID number types), which already match `or_(title, title_tr)`.

Endpoints fixed:
- `/admin/api/labels/`
- `/admin/api/sources/`
- `/admin/api/eventtypes/`
- `/admin/api/potentialviolation/`
- `/admin/api/claimedviolation/`

No schema changes. Entity search (search_utils) is unaffected since it filters by id.

Addresses the searchability half of BYNT-1761; displaying translated values in pickers is a separate follow-up.

## Test
Added `test_lookup_search_by_translated_title` covering all five endpoints (fails without the fix, passes with it). Full `tests/test_lookup_crud.py` module passes (217 passed).

### DIFF
--- a/enferno/admin/views/eventtypes.py
+++ b/enferno/admin/views/eventtypes.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from flask import Response, request
 from flask.templating import render_template
 from flask_security.decorators import current_user, roles_accepted, roles_required
+from sqlalchemy import or_
 
 from enferno.extensions import db
 from enferno.admin.constants import Constants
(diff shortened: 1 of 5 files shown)


apodacaduron (approved):
Note: sources does not have a way to add title_ar on the UI

PR #369, opened 2026-07-19
Show translated titles in lookup admin tables and label tree payload

level09:
## Summary
- Add a Title (AR) column to the Labels and Sources admin tables so imported translations are reviewable at a glance (Event Types and the component-data lists already show theirs)
- Include `title_ar`, `comments`, `comments_ar` in `Label.build_tree`: the tree view feeds the edit dialog from this payload, and without these fields saving from tree view silently wiped a label's translation and comments (`from_json` defaults them to empty)
- Add `title_ar` to `Source.to_dict` and a Title (AR) field to the source edit dialog; the column existed and is searchable (#365) but was neither visible nor editable

## Test plan
- `uv run pytest tests/test_lookup_crud.py` (217 passed)
- Manual: labels admin tree view -> edit a translated label -> Title (AR) populated -> Save -> translation preserved
- Manual: sources admin -> table shows Title (AR), edit dialog round-trips it


apodacaduron (dismissed):

apodacaduron (dismissed):

enferno/admin/models/Bulletin.py
```diff
@@ -697,19 +699,25 @@ def to_dict(self, mode: Optional[str] = None) -> dict[str, Any]:
...
+                    {"id": source.id, "title": source.title, "title_ar": source.title_ar}
+                )
 
         # labels json
         labels_json = []
         if self.labels and len(self.labels):
             for label in self.labels:
-                labels_json.append({"id": label.id, "title": label.title})
+                labels_json.append(
+                    {"id": label.id, "title": label.title, "title_ar": label.title_ar}
```
apodacaduron: Question: should we add `path` and `path_ar` here too? So we can implement the "Full label paths, without the clutter" feature

level09: @apodacaduron agreed. and pushed in f8c187fb7 

Thanks for catching this.

apodacaduron (approved):

enferno/admin/models/Bulletin.py
apodacaduron: @level09 hey, we might need `path` and `path_ar` on Actor Profile labels too. Actor profiles render labels in the actor drawer, and I think those label objects may still be missing the path fields.

PR #370, opened 2026-07-19
Add translated titles to location admin levels and location types

level09:
## Summary
- Add `title_tr` to `location_admin_level` and `location_type` (the only two lookup lists with no secondary-language support)
- Alembic migration `b6e3d1a8f254` also merges the two open migration heads on main (`68396035f041`, `c7d2e9f4a1b8`)
- Expose the field in `to_dict`/`from_json`, validation models, and the component-data admin tables (same Translated Title column and editing the other lookup lists already have)
- Typeahead endpoints for both lists now match the translated title too, consistent with #365

## Test plan
- `flask db upgrade` applies cleanly and resolves to a single head
- Component data -> Location Administrative Levels / Location Types: Translated Title column visible and editable, round-trips on save
- `GET /admin/api/location-types/?q=<arabic>` matches once translations are entered


apodacaduron:
Note: title_tr does get saved correctly but it's not rendered anywhere else in the ui

apodacaduron (approved):
Note: title_tr does get saved correctly but it's not rendered anywhere else in the ui

PR #371, opened 2026-07-19
Prevent orphan actors from interrupted create requests

level09:
## Summary
Production incident: a DB stall during actor save left an actor row committed with no `actor_profile` and no creation history. Text search inner-joins profiles, so the orphan was invisible to search while still reachable by ID; the user retried and produced a duplicate.

Two changes:
- `Actor.from_json` now processes `actor_profiles` before events/medias/relations. Those blocks call `save()` mid-flow, committing the half-built actor via session cascade; with profiles wired first, any such commit persists the actor together with its profiles, so a search-invisible actor can no longer be created.
- `api_actor_create` wraps the create flow and, on failure, deletes the partially committed actor before returning 500, so a client retry cannot leave duplicates behind.

## Test plan
- New test: `TestActorCreateAtomic` forces a failure after save and asserts no actor row survives
- `uv run pytest tests/test_core_entities.py tests/test_specialized.py` (439 passed, 4 skipped)


apodacaduron (approved):
Approved, this addresses the bug where a failed create could leave behind an actor without profiles, making it invisible in search

PR #372, opened 2026-07-19
Timeout-then-queue background search

level09:
## Summary

Long searches no longer hold a web worker (and its GIL) hostage. Interactive searches on the bulletin, actor, and incident list endpoints now run under a Postgres `statement_timeout` (default 30s, `SEARCH_TIMEOUT` env, 0 disables). When the database cancels one, the same query is re-queued to Celery, which re-runs it with a generous bound (default 10 min, `BACKGROUND_SEARCH_TIME_LIMIT`), stores up to 10k matching ids in Redis for 24 hours, and notifies the user with a link that replays the stored ids in the normal list view via the existing `ids` filter.

This is the same sync-with-async-fallback model used by Elasticsearch's async search API (`wait_for_completion_timeout`) and Superset's SQL Lab (`SQLLAB_TIMEOUT` + Celery).

- `enferno/utils/background_search.py`: timeout helper (`SET LOCAL`, so the bound dies with the transaction and can't leak through the pool), 57014 detection, Redis result store, and a `timeout_fallback` decorator that converts a canceled search into a 202 `{queued, token}` response
- `enferno/tasks/background_search.py`: Celery task; id-only select, completion/failure notification (new `BACKGROUND_SEARCH_STATUS` event, in-app only)
- The three list endpoints get the decorator plus one `apply_search_timeout()` call placed after activity logging (the activity commit would otherwise end the transaction and drop the bound)
- `GET /admin/api/background-search/<token>`: owner-only stored ids
- Frontend: 202 shows a snackbar; `?bgs=<token>` on the list pages loads stored ids and runs the search

## Test plan

- `uv run pytest tests/test_background_search.py` (timeout detection via real `pg_sleep` cancel, result store round-trip, full task run asserting stored ids + notification, endpoint 202 path)
- `tests/test_core_entities.py` and `tests/test_search_query_combination.py` pass unchanged
- Manual: set `SEARCH_TIMEOUT=1`, run a heavy search, confirm snackbar, notification, and results link


apodacaduron:
In my existing setup without recreating the project from scratch, i recieved the snackbar but not the notification
```json
{"timestamp": 1785269909.8868532, "level": "ERROR", "message": "Task enferno.tasks.background_search.background_search[5d87466f-2d1f-4efc-bcaa-4fbcc76ac9a9] raised unexpected: AttributeError(\"'NoneType' object has no attribute 'get'\")", "logger": "celery.app.trace", "pathname": ".venv/lib/python3.14/site-packages/celery/app/trace.py", "lineno": 285, "exception": {"type": "AttributeError", "message": "'NoneType' object has no attribute 'get'", "traceback": "Traceback (most recent call last):\n  File \".venv/lib/python3.14/site-packages/celery/app/trace.py\", line 479, in trace_task\n    R = retval = fun(*args, **kwargs)\n  File \"enferno/tasks/__init__.py\", line 77, in __call__\n    return super(ContextTask, self).__call__(*args, **kwargs)\n  File \".venv/lib/python3.14/site-packages/celery/app/trace.py\", line 779, in __protected_call__\n    return self.run(*args, **kwargs)\n  File \"enferno/tasks/background_search.py\", line 53, in background_search\n    Notification.send_notification_for_event(\n  File \"enferno/admin/models/Notification.py\", line 178, in send_notification_for_event\n    config = get_notification_config(event)\n  File \"enferno/admin/models/Notification.py\", line 223, in get_notification_config\n    \"enabled\": config.get(\"in_app_enabled\", True),\nAttributeError: 'NoneType' object has no attribute 'get'"}}
```
Maybe there could be a fallback? Or something like that for existing projects? Not sure

apodacaduron:
Manual testing found that the notification link renders as literal HTML. We changed notifications from `v-html` to `v-text` in the pentest fixes, so `<a href=...>` is escaped now.
Screenshot 2026-07-28 at 4 45 07 p m

level09:
Thanks @apodacaduron, both are fixed now.

You were right about the fallback. Your config.json is older than this PR, so it does not have the new event yet, and the code was crashing instead of using the defaults. Now it falls back to the defaults, so any new event we add later will not break existing installs.

For the link, I moved it out of the message into a new `link` field on the notification. The message stays plain text, so we keep the `v-text` escaping from the pentest fixes, and the link renders as a normal anchor. Added a migration for the new column.

Please run `flask db upgrade` after you pull, otherwise the notifications list will fail on the missing column.

Can you pull and test again when you have time?

apodacaduron (approved):
@level09 Yes now works as expected

PR #378, opened 2026-07-22
Add label structure navigator

level09:
## Summary

- add a searchable, read-only label tree to the app bar for Admin, Mod, and DA users
- include Arabic titles and entity availability indicators while keeping label management restricted to Admin and Mod

## Tests

- `uv run pytest -q` (867 passed, 4 skipped)


apodacaduron (dismissed):
Works well, just noticed something unrelated to this PR, im unable to edit existing Labels since it says 
```
[path]: Extra field not allowed
[path_ar]: Extra field not allowed
```

level09:
Thanks @apodacaduron good catch, addressed

apodacaduron (approved):
Hey i made some changes, fixed the fab/panel positioning, it was using a guessed pixel offset so it'd land on top of the AI/Vision badge and get pushed to a corner on resize instead of staying put. now it measures the actual toolbar to figure out where to dock (works in RTL too), and positions are anchored to the nearest edge so resizing doesn't break it. also closes the panel when you click "Manage Labels" instead of leaving it floating over the page.

PR #379, opened 2026-07-22
Show label hierarchy paths in label previews

apodacaduron:
## Summary
Labels can be nested, but pickers and chips only showed the leaf title, so same-named or generic labels (e.g. "Other") under different parents were indistinguishable. This adds path-aware label display:

- New `LabelSearchField` (extends `SearchField`) — labels-only autocomplete using the compact `mode=2` payload, showing parent path as a subtitle and a collapsed path on selected chips when the leaf is ambiguous.
- New `LabelPath.js` — path utils + `LabelPathChip`/`LabelPathTrail`/`LabelPathList`, replacing flat label chips on bulletin/incident/actor cards and profiles. RTL-aware.
- Swapped `search-field` → `label-search-field` everywhere labels/ver_labels are picked (bulletins, actors, incidents, review dialog, media import).

Note: card/profile views need a `path` field added to their serializers to actually show hierarchy — landing in a separate follow-up PR. Until then those views keep showing flat leaf chips.

Also: closes the incident drawer when editing an incident, to match bulletins/actors.

## How to Test
1. Create a nested label (e.g. parent "Weapons", child "Firearms").
2. In a labels picker (bulletin/actor/incident), search for the child — confirm the dropdown shows it with the parent as a subtitle.
3. Select two same-named/generic labels under different parents — confirm the chip shows the parent path, not just the leaf.
4. Save and reopen the entity — confirm labels persist.
5. Switch UI to Arabic and repeat 2–3 — confirm path direction/order flips correctly.
6. Open an incident from the list — confirm the edit drawer closes.
7. Check console for errors on touched pages.


enferno/static/js/components/BulletinCard.js
```diff
@@ -341,11 +341,7 @@ const BulletinCard = Vue.defineComponent({
                   <v-toolbar-title class="text-subtitle-1">{{ translations.labels_ }}</v-toolbar-title>
                 </v-toolbar>
                 <v-card-text class="pt-0">
-                  <div class="flex-chips">
-                    <v-chip label size="small" class="flex-chip" v-for="label in bulletin.labels" :key="label.id">
-                      {{ label.title }}
-                    </v-chip>
-                  </div>
+                  <label-path-list :labels="bulletin.labels"></label-path-list>
```
level09, line 348: This card is also used on pages not touched by this PR (export dashboard, deduplication, import log). They don't register `LabelPathList`, so labels there render empty. Registering it inside the card itself (`components: { LabelPathList }`) protects every page, then those 3 pages just need the script tag.

level09 (commented):
Nice work, the new picker and the path chips are a good direction. Two things break real pages, and one ordering note:

1. The cards (bulletin / incident / actor profiles) are also used on the export dashboard, deduplication and import log pages. Those pages don't load `LabelPath.js` or register `LabelPathList`, so labels will show empty there.
2. Same problem on the activity page with `label-search-field` (details inline).
3. Merge order: the Arabic data (`title_ar` / `path_ar` in mode 2) comes from PR #369, so that one should go in first, then re-test Arabic here.

enferno/admin/templates/admin/activity.html
```diff
@@ -401,6 +402,7 @@
   app.component('ActivitySearchBox', ActivitySearchBox);
 
   app.component('SearchField', SearchField);
+  app.component('LabelPathList', LabelPathList);
```
level09, line 407: This page also renders the bulletin/actor/incident search boxes, which now use `label-search-field`, but it's not loaded or registered here, so the label filters on this page will be empty.

enferno/static/js/components/LabelSearchField.js
```diff
@@ -0,0 +1,144 @@
...
+          isRtl: Boolean(window.LabelPathUtils?.isRtl()),
+          leaf: item?.title || item || '',
+          markerIcon: window.LabelPathUtils?.isRtl() ? 'mdi-chevron-left' : 'mdi-chevron-right',
+          parent: '',
+          showParent: false,
+        };
+      }
+      return window.LabelPathUtils.chipParts(item, this.selectedDuplicateLeaves);
+    },
+    debouncedSearch: debounce(function (search) {
```
level09: I think you can delete this override and `labelQueryParams`: every caller already sends `mode: 2`, so a default value for the `api` prop is enough and the base SearchField does the rest.

enferno/static/js/components/LabelPath.js
```diff
@@ -0,0 +1,343 @@
...
+    const arabic = this.splitPath(label.path_ar || label.path_tr);
+    if (!english.length) return this.parentSegments(label, arabic, 'ar');
+    if (!arabic.length) return this.parentSegments(label, english, 'en');
+
+    const length = Math.max(english.length, arabic.length);
+    const segments = Array.from({ length }, (_, index) => arabic[index] || english[index]).filter(Boolean);
+    return this.parentSegments(label, segments, 'ar');
+  },
+
+  parentSegments(label, segments = [], lang = this.primaryLang()) {
```
level09: The backend path never includes the label itself, so this trimming never triggers on real data. Except one bad case: a child label with the same title as its parent (that's allowed) would lose its path. Safer to remove it.

enferno/static/js/components/LabelPath.js
```diff
@@ -0,0 +1,343 @@
...
+  secondaryTitle(label) {
+    return this.primaryLang() === 'ar' ? (label?.title || '') : (label?.title_ar || label?.title_tr || '');
+  },
+
+  pathSegments(label, lang = this.primaryLang()) {
+    if (!label) return [];
+    const english = this.splitPath(label.path);
+    if (lang !== 'ar') return this.parentSegments(label, english, 'en');
+
+    const arabic = this.splitPath(label.path_ar || label.path_tr);
```
level09: `title_tr` / `path_tr` don't exist on labels (that's the lookups model), these fallbacks can go.

enferno/static/js/components/LabelPath.js
```diff
@@ -0,0 +1,343 @@
+const LabelPathUtils = {
+  genericLeaves: new Set(['other', 'unknown', 'أخرى', 'آخر', 'غير معروف', 'غير معروفة']),
```
level09: This word list will drift from real data ("misc", "N/A", other Arabic words...). The duplicate-leaf check already covers the ambiguous case, I'd drop the list.

enferno/static/js/components/LabelPath.js
```diff
@@ -0,0 +1,343 @@
...
+      <v-card
+        class="border rounded-lg"
+        elevation="8"
+        width="420"
+        max-width="calc(100vw - 32px)"
+        @keydown.esc.prevent="menu = false"
+      >
+        <v-card-text class="pa-3">
+          <div class="text-overline font-weight-bold text-medium-emphasis mb-2">
+            LABEL PATH
```
level09: Hardcoded English, same for "+N more" below. Should go through translations.

level09 (approved):

PR #380, opened 2026-07-25
Enable RTL layouts

level09:
Makes the admin UI render correctly right-to-left. Split out of #377, which now carries string translations only.

## Direction plumbing
- `current_language` / `current_direction` context processor derived from the active locale via `babel.Locale.character_order`
- `<html lang dir>` on all four layouts (main, login, security, setup wizard)
- Vuetify `locale` option so its own components follow suit

## Logical properties
- `location="right"` -> `location="end"` on all 11 first-party navigation drawers plus the map drawer
- `ml-*` / `mr-*` -> `ms-*` / `me-*` in the components that carried directional spacing
- `left-auto` / `right-0` split-view classes replaced with `inset-inline-start-0` / `inset-inline-end-0` utilities
- Map visualization: viewport padding, drawer toggle position/radius, chevron and back-arrow icons all mirror
- `dir="auto"` on notification title/message (user content can be either direction), `dir="ltr"` on the timestamp
- Nav-bar logo moved into the app-bar `prepend` slot with a tightened SVG viewBox, so it sits next to the menu icon in both directions

## Tests
`tests/test_rtl.py`: direction rendering for ar/en plus guards against reintroducing hardcoded `location="right"` and physical spacing in the swept files.


apodacaduron (approved):
I made some changes like the notification/avatar/language menu layering, notification panel RTL layout, entity/detail action spacing, floating visualization controls, media redactor/viewer RTL behavior, SplitView resizing, and related icon spacing issues.

Approved.

PR #381, opened 2026-07-25
Allow label edits to round-trip the path fields

level09:
`Label.to_dict()` returns `path` / `path_ar` (added in #369), and the admin edit form PUTs that payload straight back. `LabelValidationModel` is strict, so every label edit fails with:

```
[path]: Extra field not allowed
[path_ar]: Extra field not allowed
```

Adds both to the same discard list that already carries `id`, `order`, `updated_at`, `parent_label_id` and `children`.

Reported by @apodacaduron on #378.

## Tests
- New regression test PUTs a label's own `to_dict()` back and expects 200 (fails with 400 without the fix).
- `uv run pytest`: 869 passed, 4 skipped.


level09:
Folded into #378 instead, same author, and that PR is already approved.

PR #388, opened 2026-08-03
Show current Bayanat version in profile dropdown and dashboard footer

apodacaduron:
## Description
Displays the running version (config.VERSION) in the profile dropdown menu and the dashboard footer, linked to its GitHub release notes. Falls back to plain, unlinked text when the version resolves to the 0.0.0 placeholder so the link never points at a nonexistent release tag.

## How to Test
- Log in and open the profile dropdown (top-right avatar).
- Confirm Bayanat v<version> appears below Logout, linking to https://github.com/sjacorg/bayanat/releases/tag/v<version> in a new tab.
- Go to the Dashboard and confirm the same version text/link appears in the footer.

## Jira ID (if applicable)
[BYNT-1762](https://syriajustice.atlassian.net/browse/BYNT-1762)

[BYNT-1762]: https://syriajustice.atlassian.net/browse/BYNT-1762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


level09 (changes requested):
Nice and small. The plumbing was already on main (`layout.html` publishes `window.__APP_VERSION__`), so this being display only is the right call.

One thing I'd like changed before merge. When `settings.py` can't read `pyproject.toml`, `_read_version()` falls back to the literal string `"0.0.0"`. We correctly refuse to link that, but we still print it, so a broken install ends up telling every user it's running "Bayanat v0.0.0". The whole point of the ticket is answering "what version am I on", and a confidently wrong answer is worse than none. Easiest fix:

```js
appVersion() {
  const v = window.__APP_VERSION__ || '';
  return v === '0.0.0' ? '' : v;
},
```

Then `hasReleaseLink` is just `this.appVersion`, the `v-else-if` branch can go, and same for the `{% else %}` in dashboard.html.

Two other things, both your call:

The release link is `/releases/tag/v<version>`, which only resolves if the version matches a published tag here. `BAYANAT_VERSION` can be set to anything, and prod runs a branch whose version won't always match a public tag, so it can land on a 404. Pointing at `/releases` avoids that.

ProfileDropdown lives in layout.html, so the version already shows on every page. The dashboard footer then shows it a second time, through a different implementation reading `config.VERSION` instead. I'd drop the dashboard.html half and keep just the dropdown.

Rest looks good, `rel="noopener"` is there and the missing trailing newlines got cleaned up.

apodacaduron:
@level09 right! updated

level09 (approved):
@apodacaduron awesome , solid work

PR #391, opened 2026-08-06
Improve dynamic form builder clarity and field creation feedback

apodacaduron:
## Description
This PR improves the dynamic form builder UI and related admin hints for Actors, Bulletins, and Incidents.
Changes include:
- Adds a concise info hint to Dynamic Form Builder pages: “Default fields are locked. Custom fields are editable.”
- Fixes “Dynamic From Builder” typo to “Dynamic Form Builder”.
- Adds administrator-only links from Actor/Bulletin/Incident tables to their respective form builders.
- Hides those table footer hints on mobile using Vuetify responsive classes.
- Improves the Help Text editor with a short explanatory caption.
- Updates the drag handle icon to better communicate draggable fields.
- Highlights newly added fields with a temporary blue tint instead of a border/ring animation, with dark-mode support.

## How to Test
1. Open /admin/bulletin-fields/, /admin/actor-fields/, and /admin/incident-fields/.
2. Confirm the page title says “Dynamic Form Builder”.
3. Confirm the hint appears under the toolbar with an info icon: "Default fields are locked. Custom fields are editable."
4. Add a new custom field and confirm it scrolls into view with a temporary blue highlight.
5. Open a custom field’s edit mode and confirm Help Text has the explanatory caption.
6. Visit the Actor, Bulletin, and Incident table pages as an admin.
7. Confirm the footer shows the form-builder hint on desktop.

## Jira ID (if applicable)
[BYNT-1766](https://syriajustice.atlassian.net/browse/BYNT-1766)
[BYNT-1767](https://syriajustice.atlassian.net/browse/BYNT-1767)
[BYNT-1768](https://syriajustice.atlassian.net/browse/BYNT-1768)

[BYNT-1766]: https://syriajustice.atlassian.net/browse/BYNT-1766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


level09 (changes requested):
The clarity side is good. The locked vs editable caption, the admin hint and the help text description all land, and thanks for the `From Builder` typo.

One thing blocks it. The form builder templates render `filteredMovableDynamicFields`, not `dynamicFields`, and that computed filters on the search box. So if anything is in the search box when you save a new field, the field is filtered straight out of the DOM. `querySelector` finds nothing, the `?.` swallows it, and nothing happens: no scroll, no flash.

To reproduce: open Actor Form Builder, search `name`, leave it there, click Add New, title the field `zzz test`, save. Nothing happens. Clear the search and it's sitting there, unhighlighted.

That's the exact flow the ticket is about, and the search box sits right next to the Add New button, so it will happen in normal use. Clearing the search as part of adding is enough, since a new field should always be shown:

```js
this.ui.search = '';
event.justAdded = true;
this.formBuilder.dynamicFields[entityType].push(event);
```

Two smaller things, neither needs to be in this PR:

The `setTimeout` that sets `justAdded = false` writes to the raw object rather than the reactive proxy, so it never triggers the re-render that removes the class. No visible effect, since the animation ends at `opacity: 0` on its own. Easiest is to drop the timeout entirely and let the CSS finish.

Eight new strings and no `.po` touched, so Arabic falls back to English. Same situation as #303, so it's probably one `flask translate extract && update` pass covering both, with Khaleel doing the Arabic. The typo rename is free by the way, `Dynamic From Builder` was never translated, so nothing is lost.

One I checked and it's fine: the `>` in the admin hint mirrors itself in Arabic, since U+003E is bidi-mirrored, and the labels match the nav drawer exactly.

apodacaduron:
@level09 Right, now search clears when adding a new field. BTW i didn't know that command for translations existed, thanks for pointing it out. I'll run it after wrapping up translation changes going forward.

level09 (approved):

PR #395, opened 2026-08-10
Keep session alive during active typing/reading, warn before expiry

apodacaduron:
## Description
Sessions were expiring mid-sentence. Idle timeout only slid forward on
HTTP requests, so typing or scrolling (no requests sent) looked idle and
users got logged out mid-form.

Adds a global session controller, mixed into all admin pages:
- Tracks real interaction (keydown, input, pointerdown, wheel, touchmove)
  and pings /admin/api/session-check on a cadence based on the session
  lifetime. Programmatic scroll doesn't count, only wheel/touchmove.
- Shows a 60s countdown with "Stay signed in" before actual expiry.
  Only clicking it extends the session.
- Uses localStorage so multiple tabs share one keepalive and only the
  active tab shows the warning.
- Also fixes: split-view tabs never fire visibilitychange on each other,
  so a stale "Session Expired" dialog in tab B wouldn't close after
  signing in on tab A. Sign-in now broadcasts to sibling tabs.

Frontend only, no API/DB changes. Exposes the existing
PERMANENT_SESSION_LIFETIME via one Jinja attribute. Notification poller
(#343) untouched.

Not included: request recovery after expiry, draft autosave.

## How to Test
Set SESSION_LIFETIME=120 in .env, restart, revert before merge.

- Type on and off for 3-4 min: no logout, one keepalive every ~40s
- Scroll a long record with trackpad only, no keyboard: no logout
- Type then stop: warning shows around 1 min left, expires if ignored
- Click "Stay signed in": dialog closes, timer resets
- 5 tabs, use one: only that tab warns and pings
- Split view, both expired, sign in on one: other tab closes on its own
- Sleep past expiry, wake: no stale countdown
- window.scrollTo() from console with no real input: no keepalive

## Jira ID (if applicable)
[BYNT-1763](https://syriajustice.atlassian.net/browse/BYNT-1763)

[BYNT-1763]: https://syriajustice.atlassian.net/browse/BYNT-1763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


level09 (changes requested):
Nice work, the cross-tab handling especially.

One thing blocking. The warning dialog closes itself when the countdown reaches zero:

```js
this.sessionWarningVisible = remainingMs > 0 && remainingMs <= this.sessionWarningWindowMs;
```

So someone who steps away comes back to a page that looks completely normal but has a dead session, and only finds out when their next click 401s. It does recover through the 401 interceptor, but that's the least clear moment in a feature built to make expiry clear. At zero it should hand off to the sign-in dialog instead of just disappearing.

To reproduce: set `SESSION_LIFETIME=30` in .env, restart, log in, don't touch anything and watch the countdown hit zero. Dialog vanishes, page looks fine, session is gone.

apodacaduron:
@level09 Right! missed it because of the higher session lifetime i used (120s), it allowed notification poll to return 401 so i didnt noticed, fixed now

level09 (approved):
@apodacaduron very good fix, thanks

PR #397, opened 2026-08-11
Make the system users list searchable

level09:
Salvaged from the closed #217 and rebuilt on current main.

The System Users page has no way to find a user. It turns out this is not a missing feature so much as a broken one on both ends:

- The page passes `{search: this.search}` as the **second argument to `api.get()`**, which axios reads as the request *config*, not as params. An unknown `search` key there is silently dropped, and `this.search` was never defined on the component in the first place. Nothing was ever sent.
- The backend has read a `q` param all along, so `q` was dead code that nothing could reach.
- There is no search field in the toolbar.

### Changes
- Search field in the users toolbar, matching the existing pattern on the Sources page (`clearable`, `@click:clear="resetSearch"`, submit on Enter) so it behaves the way the other admin lists already do.
- `refresh()` now sends real query params via the axios `params` object.
- `q` matches name, username, or email instead of name alone. Case-insensitive partial match.

### Permission note
`name`, `username` and `email` are all masked by the `secure_*` properties for users without `view_usernames`, and `to_compact()` returns the masked values. Filtering on a masked field turns the list into an oracle: `ilike` is a substring match and the response carries a `total`, so a caller could reconstruct values a row at a time rather than merely confirming a guess. Search is therefore gated as a whole on the same condition the masking uses (`Admin` or `view_usernames`), and callers without it get no matches at all rather than a name-only fallback.

No UX cost: the users page is `@roles_required("Admin")` while only the API is open to `Mod`, so nobody loses a working search box. The unfiltered list is untouched and stays masked as before.

That condition now lives in one place, `can_view_usernames()` in `enferno/user/models.py`. It previously existed as four hand-copied variants (three `secure_*` properties plus a nested copy in `check_history_access`), only one of which handled the no-request-context case. Drift between the masking rule and the search gate is what produced the `name` gap in the first place.

`test_hidden_identifiers_are_not_searchable` is parametrized over name, username and email, and all three probes fail if the gate is removed (verified by bypassing it).

### Verification
- 898 passed, 4 skipped (full suite), across two runs.
- Guard tests confirmed to fail when the permission branch is bypassed, so they are not tautological.
- Template parses.

### Not included
Status and role filters. #217 had them, but they are a filter-row design decision on a page Daniel owns, and there is no ticket or mockup driving them yet. The search field is the unambiguous part and stands on its own. Worth raising with him against BYNT-1549.


apodacaduron (dismissed):
Working well just made and adjustment to v-model so we trim query the same way as other places

apodacaduron (approved):

PR #400, opened 2026-08-18
Production-ready Docker deployment with automatic HTTPS

level09:
Makes the Docker Compose deployment production-capable. Native installation remains the recommended path.

## Changes

- Caddy replaces the nginx edge. `DOMAIN` in `.env` enables automatic HTTPS via Let's Encrypt.
- `gen-env.sh -d` prompts for the domain and sets `SECURE_COOKIES` and `FORCE_HTTPS`. Both were previously unset, so the session cookie lacked the Secure flag behind TLS.
- Setup creates `config.json` and the data directories owned by the container user. Without this a fresh clone fails to start: Docker creates a directory in place of `config.json`, and root owned bind mounts leave the application unable to write its log file.
- Adds restart policies, container log rotation, and a health check covering the database and Redis. The edge waits for the application to be healthy.
- uWSGI runs multiple workers with a request timeout and worker recycling, overridable from `.env`.
- Build context restricted to application files with an allowlist.
- Docker deployment guide rewritten: install, upgrade, rollback, backups, teardown, troubleshooting.

## Notes

- Only the edge publishes ports. PostgreSQL and Redis stay on the internal network.
- Changes to `.env` require `docker compose up -d --force-recreate`.
- Behind an existing TLS proxy, `FORCE_HTTPS` must stay `False` or the application redirects in a loop.
- No automatic pre-upgrade snapshots on this path. The guide documents taking a dump before upgrading.
- `docker-compose-test.yml` and `flask/Dockerfile.test` are unchanged.

## Testing

Clean Ubuntu host: fresh clone install, certificate issued and persisted across restart, HTTPS login, migrations on restart, backup and rollback, `flask doctor`, full test suite (890 passed).


level09:
Closing in favour of consolidating this into the upcoming v5.0.0 release PR.

All seven commits here have been carried over as-is, so nothing is dropped: the production compose stack with the Caddy TLS edge, the health-gated startup, the bind-mount ownership and FORCE_HTTPS healthcheck fixes, the build-context allowlist, and the Docker deployment guide.

The reason for folding it in rather than merging separately is testing: v5 changes the installer, the update path and the deployment layout, and we want one branch validated end to end rather than several merged in sequence. The release PR will follow shortly.

PR #401, opened 2026-08-19
Replay session-expired requests after re-authentication

apodacaduron:
## Description
Fixes a bug where a 401 from session expiry dropped the request silently — if it was a save, the user could think it went through when it didn't.

Now the request is queued, the sign-in dialog opens, and once the user re-authenticates it replays automatically. Freshness/step-up challenges are excluded and still require a manual retry. Also fixed two role management actions (save, CSV import) that had no error handling at all, which meant a failed freshness reauth there gave no feedback. Old callbackQueue mechanism removed.

## How to Test
1. Fill a new bulletin form, before clicking save, on another tab or device logout all sessions from user, then go back to form and click save, form should return a 401 and dialog should appear, after sign in it should replay request.
2. Confirm it only saves once
3. Let freshness expire (System Administration save), verify the "Verify Password" dialog shows and the save does NOT auto-replay — must click Save again.
4. Setup wizard still works end to end.

## Jira ID (if applicable)
[BYNT-1764](https://syriajustice.atlassian.net/browse/BYNT-1764)

[BYNT-1764]: https://syriajustice.atlassian.net/browse/BYNT-1764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/enferno/admin/templates/admin/roles.html
+++ b/enferno/admin/templates/admin/roles.html
@@ -280,7 +280,9 @@
                         this.imDialog = false;
                         this.refresh();
                         this.showSnack(response.data);
-                    })
+                    }).catch(error => {
+                        this.showSnack(error?.response?.data);
+                    });
                 },
 
                 newItem() {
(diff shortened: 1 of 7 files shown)


level09:
Thanks for working through this Daniel. Looking at it again, I think I overcomplicated the ticket.

We are in a good place already. The missing catch handlers in `roles.html`, the `resetGroupFormState` rename and retiring `callbackQueue` are the parts that fix real bugs.

The replay queue is more machinery than the problem needs. Once the sign in dialog appears the user can see the save did not go through, and the form is still there, so it costs them a click rather than losing work.

Can we go with the simpler version? Either strip this PR down to those parts, or close it and open a small one. Your call.

PR #402, opened 2026-08-20
chore(deps): upgrade Python dependencies and enable dependabot

level09:
## Summary

General dependency refresh off `main`, plus Dependabot so this stops being a manual chore.

## Changes

- `uv.lock`: full `uv lock --upgrade`. 95 packages updated, 4 added, 7 removed.
- `pyproject.toml`: raise the `yt-dlp` floor to `>=2026.8.19`.
- `.github/dependabot.yml`: weekly `uv` and `github-actions` updates, minor/patch grouped into a single PR to avoid noise. `flask-security-too`, `black` and `ruff` are ignored since they are pinned deliberately.
- `docs/guide/web-import.md`: document the JavaScript runtime requirement for web import.

## Why the yt-dlp bump matters

YouTube now serves media behind a JavaScript challenge. yt-dlp needs a JavaScript runtime (Deno by default) available on the server to solve it, plus a recent release to handle current streaming behaviour. An older yt-dlp fails with `HTTP Error 403: Forbidden` even when the proxy and cookies are configured correctly.

Because large platforms change extraction logic frequently, a stale yt-dlp is a common cause of web imports that used to work suddenly failing. The docs now cover installing a JavaScript runtime alongside keeping yt-dlp current.

## Notable major bumps

`webauthn 2.7.0 -> 3.0.0`, `redis 7.1.0 -> 8.1.0`, `cryptography 49 -> 50`, `chardet 5 -> 7`, `weasyprint 68 -> 69`, `cbor2 5 -> 6`, `more-itertools 10 -> 11`, `rich 14 -> 15`, `zipp 3 -> 4`.

## Testing

Full suite run before and after the upgrade with identical results: **884 passed, 4 skipped**. The only failures in either run were 2 pre-existing ones (`test_bay_01_025_pdf_url_fetcher_*`) that fail on macOS because pango/gobject are absent, and pass in CI. No new failures introduced.

Upgrading requires `uv sync` and a restart of the app and Celery worker.


apodacaduron (approved):
Approved. I tested the main risk areas from the dependency refresh: uv sync plus app/Celery restart, session flow, Celery/background tasks, YouTube web import, 2FA/passkey flows, PDF export, media preview/OCR display, user email flow, and the main search/entity pages. Everything looked good to me. I only had to install ffmpeg locally for YouTube import, but that was an environment setup issue.

PR #405, opened 2026-08-22
chore: use the pymupdf import name instead of the deprecated fitz alias

level09:
## Summary

PyMuPDF emits a deprecation warning on every import:

```
warning: The `fitz` API is deprecated and will be removed in future. Use `import pymupdf` instead.
```

`fitz` is the legacy alias for the same package. The API is identical, so this is a pure rename: `import fitz` becomes `import pymupdf`, and `fitz.X` becomes `pymupdf.X`. 15 references across 4 files, no behaviour change.

`pymupdf` is already a direct dependency in `pyproject.toml`, so nothing new is being pulled in. We were only reaching it through the deprecated alias.

## Why now

The alias is slated for removal upstream, at which point every one of these imports breaks. Doing it while it is a mechanical find and replace is cheaper than doing it under a failing build.

## Testing

- `tests/admin/test_redaction.py`: 9 passed. These exercise the actual PDF paths (`open`, `Rect`, page redaction), so the rename is covered rather than just imported.
- Full suite matches baseline `main`: same pre-existing failures, none new.
- Confirmed the warning is gone by importing both touched modules with `warnings.simplefilter("error")`.


apodacaduron (approved):
OCR and Redactions still work fine for pdfs

PR #406, opened 2026-08-22
chore: enable CodeRabbit automatic reviews with vendored paths excluded

level09:
## Summary

Turns CodeRabbit's automatic PR reviews on, and moves the review policy into the repo instead of the dashboard.

CodeRabbit is already installed here and free for public repositories, but auto review is switched off in their UI, so it only ran when someone remembered to invoke it manually. In practice that meant almost never.

## Why the config lives here

Keeping it in `.coderabbit.yaml` makes the review policy versioned and reviewable like any other config, rather than a setting one person changed in a web UI that nobody else can see.

## Choices

- `profile: chill` - balanced. `assertive` is noticeably nitpicky on a codebase this size, `quiet` drops too much to justify the round trip.
- `request_changes_workflow: false` - advisory only. This must never gate a merge: it is non-deterministic, and a required check that can fail at random is worse than no check at all.
- `auto_review.base_branches: [main]`, drafts excluded.

## Path filters

Vendored third-party JavaScript is excluded: TinyMCE, Video.js, Lightgallery, jsondiffpatch, Dropzone, PDF.js, Croppr, qrcodejs, JSZip, docx-preview, the loose Leaflet and Vue bundles, and any `.min.js` / `.min.mjs`.

We never hand-edit those files. They change only by wholesale upgrade, so there is nothing for a reviewer to comment on, and library CVEs are retire.js's job. That leaves review focused on `components/`, `mixins/`, `common/` and `utils/`, which is the JavaScript we actually write.

Also excluded: translation catalogs, `uv.lock`, `docs/package-lock.json`, and Alembic revisions, all machine-generated.

## Notes

This is advisory tooling, not a security control, and it does not change any required check or replace anything currently running.

## Summary by CodeRabbit

* **Chores**
  * Added repository-level automated code review configuration.
  * Configured non-blocking reviews for eligible changes targeting the main branch.
  * Excluded generated, vendored, localized, dependency-lock, and migration files from automated review.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,49 @@
+# CodeRabbit configuration
+# Docs: https://docs.coderabbit.ai/reference/yaml-template
+#
+# Kept in the repo rather than the CodeRabbit dashboard so the review policy is
+# versioned and reviewable like any other config.
+
+language: en-US
+
+reviews:
+  # Balanced feedback. "assertive" is noticeably nitpicky on a codebase this
+  # size; "quiet" drops too much to be worth the round trip.
+  profile: chill
+
+  # Advisory only. This must never gate a merge: it is non-deterministic, and a
+  # required check that an LLM can fail at random is worse than no check.
+  request_changes_workflow: false
+
+  auto_review:
+    enabled: true
+    drafts: false
+    base_branches:
+      - main
+
+  path_filters:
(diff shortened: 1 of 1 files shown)


level09:
Closing for now. Not worth the extra review surface at the moment; CodeRabbit stays available on manual trigger. Branch kept if we want to revisit.

